PR #5, opened 2020-02-21
Added getElementBehaviorType method

A-Gray-94:
I'm implementing a geometrically nonlinear version of RLT and as a result I need to be able to tell whether an element is geometrically linear or nonlinear so I added a getter method for the `ElementBehaviorType` property of an element in the same style as the current `getElementType` method.

I've added `NONE` as a new `ElementBehaviorType` option which is the default output of `getElementBehaviorType` and each child element class would then need to redefine it's own `getElementBehaviorType` method. So far I have only done this for the MITCShell class.

Please let me know if there's a better way of implementing this.

### DIFF
--- a/src/elements/MITCShell.h
+++ b/src/elements/MITCShell.h
@@ -72,6 +72,7 @@ class MITCShell : public TACSShell {
   // ------------------------------------
   int numNodes();
   int numVariables();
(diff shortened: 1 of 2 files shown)


A-Gray-94:
Closing as I'm not working on this anymore and it's probably hopelessly out of date with all the recent development

PR #27, opened 2021-09-24
Adding shell integration tests

timryanb:


gjkennedy (commented):
Please use 2 spaces to indent all C++ code and 4 spaces to indent python code. We should implement a linter here, but try and make the code look as similar as possible to the existing TACS code.

Also, I think that the getPointwiseQuantity code should accept a NULL detXd and/or quantity argument, in which case it should only return the number of entries returned by quantity. This is an existing problem in the shell that I can address later.

gjkennedy (approved):

PR #36, opened 2021-10-06
adding tractions methods to cython and unit tests

timryanb:


src/elements/TACSElement.h
```diff
@@ -183,6 +183,15 @@ class TACSElement : public TACSObject {
     return NULL;
   }
 
+  /**
+    Create element traction class
+
+    @return The TACSElement traction class associated with this element. Possibly NULL.
+  */
+  virtual TACSElement* createElementTraction(TacsScalar t[], int faceIndex){
```
gjkennedy, line 191: I think the faceIndex argument should come first to be consistent with the way we've typically written these function prototypes.

src/elements/TACSElement2D.cpp
```diff
@@ -51,6 +52,11 @@ TACSElementModel* TACSElement2D::getElementModel(){
   return model;
 }
 
+TACSElement* TACSElement2D::createElementTraction(TacsScalar t[], int faceIndex){
```
gjkennedy, line 55: Please put a space before the first variable type and after the last argument

gjkennedy: Same throughout

src/TACSAssembler.cpp
```diff
@@ -4898,21 +4898,27 @@ void TACSAssembler::addAdjointResProducts( TacsScalar scale,
     }
 
     // Add the contribution from the auxiliary elements
-    while (aux_count < naux && aux[aux_count].num == i){
-      // Get the design variables for this element
-      numDVs = aux[aux_count].elem->getDesignVarNums(i, maxDVs, dvNums);
+    if (aux_count < naux){
```
gjkennedy, line 4901: Thanks for catching this!

src/elements/TACSElement.h
timryanb: Should be fixed now

gjkennedy (approved):

PR #43, opened 2021-10-15
Adding mpi tests to all unit tests

timryanb:
Fixed broken parallel tests


src/constitutive/TACSIsoShellConstitutive.cpp
```diff
@@ -321,40 +329,49 @@ TacsScalar TACSIsoShellConstitutive::evalFailureStrainSens( int elemIndex,
...
+    ksFactor = exp(ksWeight*(bottom - ksMax)) / ksSum;
 
-      sens[6] = sens[7] = sens[8] = 0.0;
+    sens[0] += ksFactor*phi[0];
+    sens[1] += ksFactor*phi[1];
+    sens[2] += ksFactor*phi[2];
 
-      return bottom;
-    }
+    sens[3] += ksFactor*-ht*phi[0];
```
gjkennedy, line 368: the *-ht* is a bit confusing here.. can we just make it
sens[3] -= ksFactor*ht*psi[0];

src/elements/TACSElement.cpp
```diff
@@ -233,9 +245,23 @@ void TACSElement::addAdjResProduct( int elemIndex, double time,
       product += psi[i]*TacsImagPart(tmp[i])/dh;
     }
 #else
-    for ( int i = 0; i < nvars; i++ ){
-      product += psi[i]*(tmp[i] - res[i])/dh;
-    }
+      if (fdOrder < 2){
```
gjkennedy, line 248: Indentation looks off here? 2-spaces only?

src/elements/TACSElement.cpp
```diff
@@ -20,6 +20,18 @@
...
 
+/*
+  Default finite difference order for real analysis
+*/
+int TACSElement::fdOrder = 2;
+
+/*
+  Allow users to set default finite difference order for real analysis
+*/
+void TACSElement::setFDOrder( int order ){
```
gjkennedy, line 31: what about the full name setFiniteDifferenceOrder()?

src/elements/TACSElement.cpp
timryanb: Should be addressed in last commit

src/constitutive/TACSIsoShellConstitutive.cpp
timryanb: Should be addressed in last commit

gjkennedy (approved):

PR #84, opened 2022-03-03
Moving deprecated examples into separate folder

timryanb:
- Addresses issue #1


gjkennedy:
These examples aren't "deprecated". We should fix them.

gjkennedy:
Didn't mean to close this all the way. We could organize the examples differently. Remove some of those old ones that are duplicative. But I think "deprecated" is definitely the wrong word.

timryanb:
Sorry "deprecated" might not be the best word. The idea wasn't to remove them permanently, but to make it clear for users that these examples aren't working yet. We could always move them back with the main examples as they're updated. Maybe we can add a README to each of these examples noting that they're not ready yet, if you'd prefer that?

PR #86, opened 2022-03-08
added interface to set initial conditions with a TACSBVec or, if usin…

markleader:
…g pytacs, with a numpy array or float; added an integration test for initial conditions interface


src/TACSAssembler.cpp
```diff
@@ -3816,9 +3826,11 @@ TACSSerialPivotMat *TACSAssembler::createSerialMat(){
...
                                        TACSBVec *dvars,
                                        TACSBVec *ddvars ){
-  if (vars){ vars->zeroEntries(); }
-  if (dvars){ dvars->zeroEntries(); }
-  if (ddvars){ ddvars->zeroEntries(); }
+  // If the user requests initial conditions of vars and vars0 has not been set,
+  // zero the vector to prepare to get initial conditions from the element interface
+  if (vars && !vars0){ vars->zeroEntries(); }
+  if (dvars && !dvars0){ dvars->zeroEntries(); }
+  if (ddvars && !ddvars0){ ddvars->zeroEntries(); }
```
gjkennedy, line 3834: There should be cases where you can skip this loop over the elements. For instance, if vars, dvars, ddvars are passed in and all vars0, dvars0 and ddvars0 are defined, then the loop has no effect.

gjkennedy: To be clear, I'm looking at the loop beginning at 3840 - this is executed every time, right? Or am I missing something

markleader: I see what you're saying. I'll update this

gjkennedy: Perfect! Thanks

gjkennedy (approved):

PR #119, opened 2022-06-17
DIRK Time Integration Bug Fix

bradenfrigoletto24:
### Overview
The DIRK time integrator currently implemented does not give the correct convergence behavior when used with a time-varying load. A test case and supporting scripts are attached to demonstrate the DIRK integrator performance before and after the fix.

### Test Case
The test case features a cantilevered beam with a time-varying out-of-plane tip load. Varying the time-step across multiple trials and selecting the most refined as the reference, the relative error between the trials can be evaluated. This relative error should represent the time-discretization error only since the spatial discretization error should be constant for each case. The results will show that the rate of convergence for each of the DIRK implementations is limited to about 1. After the fix is applied and tested, the results will show the proper convergence rates for the DIRK methods. The 2nd-order DIRK shows near perfect behavior, the 3rd-order DIRK shows slightly better performance than expected, and the 4th-order DIRK shows a little less performance than expected but is still high order.

[Beam_Test_Case.zip](https://github.com/smdogroup/tacs/files/8931648/Beam_Test_Case.zip)

### The Bug + Fix
The implementation of `iterate()` in class `TACSDIRKIntegrator` assumes that the forces applied are constant during the entire time step. However, since DIRK is a multi-stage method, a time-varying load should be updated at each time stage within the time step. A new class method, `iterateStage()`, is created that simply breaks down the integration logic of `iterate()` for a single time stage. This allows a new force vector to be specified for each stage of every time step, which shows the correct convergence behavior when tested. This change also requires adjustments to `TACS.pyx` and `TACS.pxd` to reflect the new class methods. Additionally, the `TransientProblem` class needed to be updated to account for the difference between a multi-step method like BDF and a multi-stage method like DIRK.

### DIFF
--- a/src/TACSIntegrator.cpp
+++ b/src/TACSIntegrator.cpp
@@ -2294,6 +2294,116 @@ int TACSDIRKIntegrator::iterate( int k, TACSBVec *forces ){
 }
 
 /*
+   Integration logic for DIRK, on a per-stage basis. Use this function to 
+   march in time when the forces change in time. This is necessary since
+   the forces need to be upated or evaluated for each stage of every 
+   time step for the order of accuracy of the method to be maintained.
+   The states for the timesteps are saved to the q, qdot, qddot variables,
+   while the stage states are saved to the qS, qdotS, and qddotS variables.
+ */
+int TACSDIRKIntegrator::iterateStage( int k, int s, TACSBVec *forces)
+{
+  // For the 0th time step, get the initial conditions and solve for the 
+  // unknown states
+  if (k == 0)
+  {
+    // Output the results at the initial condition if configured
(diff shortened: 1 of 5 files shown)


timryanb:
Good work testing and fixing the DIRK implementation @bradenfrigoletto24

I made some minor additions to your PR:
    - added a few new API calls to make it easier for users to query time stage information from the transient problem in addition to the time steps. 
    - Made your example into test case and added it to the integration test suite

timryanb (approved):

bradenfrigoletto24:
@timryanb Thanks for the review and improvements. I'm happy to contribute!

PR #128, opened 2022-09-09
Applied black to python files and clang-format to c++

gjkennedy:
Added hooks to enforce black formatting for python and clang-format with style=Google for c++. This is designed to enforce style consistency from all contributors.


A-CGray:
Just a note in case you didn't already know, once you guys merge this, you can add the squashed commit hash to a `.git-blame-ignore-revs` file ([like this](https://github.com/mdolab/adflow/blob/main/.git-blame-ignore-revs)) so that it's ignored from the git blame for the repo, doesn't work on GitHub yet but it does locally. We typically do this after applying big formatting changes, otherwise, the git blame becomes pretty useless

timryanb (approved):

PR #129, opened 2022-09-09
Switching to debug compilation to speedup CI

timryanb:


timryanb:
Looks like we shave off 5 minutes or so on the complex run @gjkennedy

gjkennedy (approved):

PR #132, opened 2022-09-28
ESDIRK time integration

bradenfrigoletto24:
-ESDIRK time integration added to TACSIntegrator and python interface

-ESDIRK adjoint not yet implemented

-Extended transientProblem to be used for multidisciplinary coupling

ESDIRK is very similar to the existing DIRK method, with the difference that the first stage is explicit, and the remaining stages are implicit. This slightly changes the integration logic and steps that need to be followed to iterate the time-marching solution. The ESDIRK method implemented includes a 4-stage, 3rd-order accurate scheme, a 6-stage, 4th-order scheme, and an 8-stage, 5th-order accurate scheme. The coefficients for these schemes are taken from [this reference](https://www.sciencedirect.com/science/article/pii/S0168927402001381). These ESDIRK schemes have been used for tightly coupled, high-order time accurate fluid-structure interaction applications. 

The transientProblem Python class was extended to allow the time integration in TACS to be executed one time instance at a time instead of iterating through the entire problem all at once. This allows structural states to be extracted from TACS and used for coupled multidisciplinary applications.

Attached is simple transient test case of a cantilevered beam with a time-varying tip load which verifies the ESDIRK time integrator order of accuracy. 

[ESDIRK_beam_test.zip](https://github.com/smdogroup/tacs/files/9668178/ESDIRK_beam_test.zip)

### DIFF
--- a/src/TACSIntegrator.cpp
+++ b/src/TACSIntegrator.cpp
@@ -2724,3 +2724,782 @@ double TACSDIRKIntegrator::getStageStates(int step_num, int stage_num,
     return tS;
   }
 }
+
+
+/*
+  constructor for TACSESDIRKIntegrator
+
+Input: 
+
+_tacs: 		the TACS assembler
+_tinit:		the initial time
+_tfinal:		the final time
+_num_steps: 	the number of time steps to take
+_num_stages: 	the number of Runge-Kutta stages between steps
+*/
+TACSESDIRKIntegrator::TACSESDIRKIntegrator( TACSAssembler * _tacs,
+                                            double _tinit, 
+                                            double _tfinal,
+                                            double _num_steps,
+                                            int _num_stages )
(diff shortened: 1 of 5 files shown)


gjkennedy:
Hi Braden, thanks for the contribution! We're using black as a formatter for python files and clang-format for C++ files (with the Google style option). Looks like those tests failed so you'll have to fix those before we can merge.

timryanb:
Great work @bradenfrigoletto24

bradenfrigoletto24:
Thanks guys. Sorry about the formatting, I didn't realize that you had started that. I'll look into reformatting my changes soon. Does Black apply to Cython as well? I did have to edit TACS.pyx/.pxd

gjkennedy:
No problem about the formatting - we haven't documented it properly yet. There's no formatter for the .pyx or .pxd files at this point.

gjkennedy (approved):

PR #139, opened 2022-11-03
Add ability to scale loads

A-CGray:
This PR adds the ability to scale the contribution of auxiliary elements to the residual and Jacobian. Useful for scaling external forcing terms

## Changes:
- added a `lambda` argument to the below TACSAssembler methods. The argument has a default value of 1 (to maintain backwards compatibility). If the value of lambda is not 1, the contributions from the aux elements are added to a separate array before being scaled and added to the output:
  - `assembleRes` 
  - `assembleRes_thread` 
  - `assembleJacobian` 
  -  `assembleJacobian_thread`
  - `assembleMatType` 
  - `assembleMatType_thread`
  - `assembleMatCombo` 
  - `addJacobianVecProduct` 
  - `addAdjointResProducts` 
  - `addAdjointResXptSensProducts`
- added a `_loadScale` attribute to the `StaticProblem` class, the user can set this value by calling `problem.setLoadScale(value)` or `problem.loadScale = value`. The second option just calls `setLoadScale` behind the scenes.
- Added a `getJacobian` method to the `StaticProblem` class that returns the stiffness matrix as a scipy sparse matrix
- Updated the `getResidual` method to incorporate the load scale and remove some redundant operations (there was an unnecessary additional call to `assembleRes`)
- made the `res` argument passed to `getResidual` an optional argument to avoid needlessly copying `self.res` into itself
- Added tests for the scaling of the residual and Jacobian
- Added displacement effects to `ShellCentrifugalTraction` so that centrifugal force is computed based on the deformed shape (and thus has a non-zero Jacobian), this is disabled by default for backwards compatibility and can be enabled using the `firstOrder` argument


timryanb:
I think this approach seems reasonable to me. 

I just made a minor modification to the getResidual method, I was thinking it makes sense to keep `res` a required argument, since this is a "get" method.

On adding a loadscale to the other problems, I'm not sure yet. Let's implement it on the StaticProblem first and see how generalizable it is. I think the loadscale method would also have to be added to the TACSIntegrator and BucklingAnalysis classes API as well, if we decide to go that route, which may be another can of worms...

Out of curiosity, why would the loadscale feature need to be added to the getMatType family of methods? I don't believe these return residuals

A-CGray:
> Out of curiosity, why would the loadscale feature need to be added to the getMatType family of methods? I don't believe these return residuals

In general, auxiliary elements can have non-zero jacobians (e.g a follower force), so if you're scaling the aux element's contributions to the residual I figured you'd also need to scale their jacobian contributions to keep the jacobian and residual consistent with each other

src/TACSAssembler_thread.cpp
```diff
@@ -101,11 +102,27 @@ void *TACSAssembler::assembleRes_thread(void *t) {
...
 
-      // Add the residual from the auxiliary elements
-      while (aux_count < naux && aux[aux_count].num == elemIndex) {
-        aux[aux_count].elem->addResidual(elemIndex, assembler->time, elemXpts,
-                                         vars, dvars, ddvars, elemRes);
-        aux_count++;
+      // Add the residual from any auxiliary elements, if the load factor is 1
+      // they can be added straight to the elemRes, otherwise they need to be
+      // scaled first
+      if (lambda == 1.0) {
```
gjkennedy, line 108: This should be: if (TacsRealPart(lambda) == 1.0) {
Won't work in complex mode

src/TACSAssembler.cpp
```diff
@@ -4126,7 +4126,7 @@ void TACSAssembler::evalEnergies(TacsScalar *Te, TacsScalar *Pe) {
 
   @param residual The residual vector
```
gjkennedy, line 4127: Please add a comment about this variable

gjkennedy (commented):
A few minor comments about the code.
Also, we'll need to be consistent on the derivatives w.r.t. nodal coordinates and material variables. Shouldn't be too hard - these already contain scaling factors (alpha/scale) as input

src/TACSAssembler_thread.cpp
timryanb: Alternatively, we could make lambda a `double`. Is there a reason we'd need to complex step over it?

A-CGray: This is true, I did think you could complex step the load factor to extract the aux element part of the residual, but there are probably better ways to do that. I will switch it back to `double` to be consistent with the `alpha`, `beta`, `gamma` scaling factors

A-CGray: nvm, I was wrong, `alpha`, `beta` and `gamma` are in fact TacsScalars, so I'll keep lambda as a TacsScalar too

A-CGray:
@gjkennedy are there any aux element types currently implemented that have non-zero Jacobians that I can test?

gjkennedy:
There aren't any classic aux-type elements that have a Jacobian. You could test this though by adding a regular element to the aux elements.

timryanb:
@A-CGray we could easily modify the centrifugal traction classes to include the node displacements in their formulation (ie  omega x omega x (r + u)). 

This would lead to an auxiliary element with nonzero jacobian

tests/integration_tests/test_shell_plate_rotated.py
```diff
@@ -149,3 +149,32 @@ def elem_call_back(
...
+    def test_residual_scaling(self):
+        """Test that the load scaling is working correctly for the point and pressure loads."""
+        for problem in self.tacs_probs:
+            with self.subTest(problem=problem.name):
+                np.random.seed(1)
+                # Check that the residual is zero if the states and load scale are both zero
+                problem.loadScale = 0.0
+                problem.u.zeroEntries()
+                problem.assembler.setVariables(problem.u)
+                problem.getResidual(problem.res)
```
timryanb, line 162: Could we simplify this down by adding `getJacobian`/`setLoadScale`/`addLoadScale` methods?

This portion of the test could look something like:
```
res = self.fea_assembler.createVec()
problem.setLoadScale(0.0)
problem.zeroVariables()
problem.getResidual(res)
```

timryanb:
Other than the above, it's looking good to me.

Probably want to run the c++ changes by @gjkennedy when you're ready

tests/integration_tests/test_shell_plate_rotated.py
A-CGray: Good idea, I implemented `getJacobian` and `setLoadScale`, what would `addLoadScale` do?

timryanb: Sorry not sure where that last one came from, I meant `getLoadScale`

timryanb: Yes, lets make the `getJacobian` logic similar to whats in `getResidual`. Also could you return a copy of `self.K` rather than the actual pointer to `self.K`. I also think that by default it should return it as a scipy matrix, but could have an option to return a TACS matrix class

timryanb: I'm thinking something like this:

```
def getJacobian(self):
        """Get the problem's Jacobian in sciPy sparse matrix format
        Returns
        -------
        tuple of 2 or 4 scipy.sparse.bsr_matrices
            A tuple of 2 scipy.sparse.bsr_matrices (A, B) if Jacobian is a TACSParallelMat, or 4
            scipy.sparse.bsr_matrices (A, B, C, D) if Jacobian is a TACSSchurMat
        """
        # Make sure assembler variables are up-to-date
        self._updateAssemblerVars()
        # Make sure stiffness mat is up-to-date
        self._initializeSolve()
        # Return copy of scipy mat
        return copy.deepcopy(self.K.getMat())
```

A-CGray: Sure, this is probably something for a later PR, but I was thinking it'd be good to split up the code that updates the stiffness matrix from the code that factors the PC. There are cases like this, and sometimes in a nonlinear solve where you just want to update the stiffness without doing the expensive refactorisation.

timryanb: No that definitely makes sense, let's include that split with the eventual PR for the nonlinear solver

tests/integration_tests/test_shell_hemisphere_mixed.py
```diff
@@ -199,3 +218,46 @@ def elem_call_back(
...
+        We test that J(loadscale=1) - J(loadscale=0) = 2*(J(loadscale=0.5) - J(loadscale=0))
+        """
+        for prob in self.tacs_probs:
+            if prob.name != "Centrifugal_firstOrder":
+                continue
+            prob._updateAssemblerVars()
+
+            prob.loadScale = 0.0
+            prob._factorOnNext = True
+            prob._initializeSolve()
```
timryanb, line 235: Just replace these lines (happens in a couple other tests) with `prob.setLoadScale(0.0)`.  Everything else looks good to me

tests/integration_tests/test_shell_plate_rotated.py
```diff
@@ -149,3 +149,30 @@ def elem_call_back(
...
+
+    def test_residual_scaling(self):
+        """Test that the load scaling is working correctly for the point and pressure loads."""
+        res = self.fea_assembler.createVec(asBVec=True)
+        scaledRes = self.fea_assembler.createVec(asBVec=True)
+        for problem in self.tacs_probs:
+            with self.subTest(problem=problem.name):
+                np.random.seed(1)
+                # Check that the residual is zero if the states and load scale are both zero
+                problem.loadScale = 0.0
```
timryanb, line 161: Same here

tests/integration_tests/test_shell_hemisphere_mixed.py
A-CGray: I actually wanted to keep the `prob.loadScale = value` form for one of these, to test that this method of setting the loadFactor works too.

A-CGray: Setting `prob._factorOnNext = True` is necessary here, otherwise the stiffness matrix isn't updated during `_initializeSolve`

timryanb: Ah, I didn't realize you made `prob.loadScale` a setter function. I'm alright with it then. Can we move `_factorOnNext = True` into the `prob.setLoadScale` method then so the user doesn't have to update it?

timryanb (approved):

timryanb:
Looks good to me, lets see what @gjkennedy thinks

tests/integration_tests/test_shell_hemisphere_mixed.py
```diff
@@ -199,3 +218,46 @@ def elem_call_back(
...
+    def test_jacobian_scaling(self):
+        """Test that the Jacobian contribution from the first order centrifugal load elements is scaled correctly by the
+        loadScale parameter.
+
+        We test that J(loadscale=1) - J(loadscale=0) = 2*(J(loadscale=0.5) - J(loadscale=0))
+        """
+        for prob in self.tacs_probs:
+            if prob.name != "Centrifugal_firstOrder":
+                continue
+            prob._updateAssemblerVars()
```
timryanb, line 231: Could we get rid of this line too?

tests/integration_tests/test_shell_hemisphere_mixed.py
```diff
@@ -199,3 +218,40 @@ def elem_call_back(
...
```
timryanb, line 231: Could we get rid of this line too?

A-CGray:
@gjkennedy Are you happy to merge this?

src/TACSAssembler.cpp
```diff
@@ -4171,6 +4173,15 @@ void TACSAssembler::assembleRes(TACSBVec *residual) {
       naux = auxElements->getAuxElements(&aux);
     }
 
+    // To avoid allocating memory inside the element loop, make the aux element
+    // contribution array big enough for the largest element
+    int maxNVar = this->maxElementSize;
+    TacsScalar *auxElemRes;
```
gjkennedy, line 4179: Set to NULL

src/TACSAssembler.cpp
```diff
@@ -4171,6 +4173,15 @@ void TACSAssembler::assembleRes(TACSBVec *residual) {
...
+    bool scaleAux = lambda != TacsScalar(1.0) && naux > 0;
```
gjkennedy, line 4180: Assign using an if condition? We don't use this type of construct elsewhere in TACS.

src/TACSAssembler.cpp
```diff
@@ -4380,20 +4416,61 @@ void TACSAssembler::assembleMatType(ElementMatrixType matType, TACSMat *A,
...
+    int naux = 0, aux_count = 0;
+    TACSAuxElem *aux = NULL;
+    if (auxElements) {
+      naux = auxElements->getAuxElements(&aux);
+    }
+
+    // To avoid allocating memory inside the element loop, make the aux element
+    // contribution mat big enough for the largest element
+    int maxNVar = this->maxElementSize;
+    TacsScalar *auxElemMat;
```
gjkennedy, line 4429: TacsScalar *auxElemMat = NULL;

gjkennedy (changes requested):
Ali - I think it'd be good to make a few minor changes to keep consistency. Otherwise looks good to me

src/TACSAssembler.cpp
A-CGray: I do this because this condition is checked in multiple places in the function, and it avoids having to check the same condition repeatedly inside the loop over all elements. Do you have a preffered way of doing this?

src/TACSAssembler.cpp
A-CGray: Done

src/TACSAssembler.cpp
A-CGray: Done

src/TACSAssembler.cpp
timryanb: I think he means something like this:

```
bool scaleAux = False; 
if (lambda != TacsScalar(1.0) && naux > 0){
  scaleAux = True;
}
```

gjkennedy (approved):

PR #145, opened 2022-11-16
added getNodesForCompIDs() to pytacs

bradenfrigoletto24:
- allows user to get the unique node IDs belonging to any component

My use case for this function is for coupling TACS with a CFD solver for FSI. For example, with a wingbox structure I do not want to include the interior nodes of the wingbox ribs when transferring loads and displacements in the coupled solution. This function allows me to only use the structural nodes belonging to the skins and LE/TE spars for the coupling. I thought this feature might be useful to others. Just FYI, I believe it should work in parallel but so far I have only worked with TACS in serial.


tacs/pytacs.py
```diff
@@ -513,6 +513,43 @@ def getCompNames(self, compIDs=None):
 
         return compDescripts
 
+    def getNodesForCompIDs(self, compIDs=None):
```
timryanb, line 516: This is returning the local node ordering (how the nodes are split across procs) right? 

If so, can we change this function name to: `getLocalNodeIDsForComps`?

timryanb:
Also, could we add this API to one of the test cases to make sure its covered under our test suite?

I think we could append as a new problem in this integration test: https://github.com/smdogroup/tacs/blob/master/tests/integration_tests/test_shell_plate_partitioned.py

Maybe you could add a case where you use the `getLocalNodeIDsForComps` to select a portion of the nodes on the plate and use `addLoadsToNodes` to then add a load to those subset of nodes?

timryanb:
Other than that, this seems like a really useful addition. Thanks @bradenfrigoletto24

tacs/pytacs.py
bradenfrigoletto24: I believe it actually returns the global node numbering. It was my understanding that TACS only used global numbering for nodes and doesn't renumber when partitioning occurs. However, maybe I am confusing that with a previous version of the code or something else entirely? 

If you can think of an easier way to directly get the local node ordering, then I can update the function. Local numbering would probably be easier to work with. The way I have been doing that on my own requires calling `getOwnerRange()` from the assembler and using the rank get the local numbering. But that conversion process I have been doing outside of `getNodesForCompIDs()`. If it makes more sense to do that in the body of the function, then I can add it there.

bradenfrigoletto24:
Just added an option to the function to return the node IDs in either their global or local ordering. I tested it in serial and parallel on a simple structure with two components and it worked well.

timryanb:
@bradenfrigoletto24, I just made a PR to your fork with some minor changes. I split up your local/global implementation and moved it into the meshloader class to be consistent with the `getLocalElementIDsFromComps` methods that exist there. The API should still be callable from pytacs like you had before, so hopefully it still works for you.

Let me know what you think

timryanb (approved):

PR #169, opened 2023-01-19
Fixing issue in AssembleMatType procedure

timryanb:
- Our nightly benchmark tests caught a change in the modal analysis results of one of our pytacs [examples](https://github.com/smdogroup/tacs/blob/master/examples/plate/analysis.py) (benchmarks can also be run locally using `$testflo -b .`)
- I narrowed the problem down to this [line](https://github.com/smdogroup/tacs/blob/master/src/TACSAssembler.cpp#L4452) of code in the `assembleMatType` procedure which was causing TACSFrequencyAnalysis to behave differently
- The problem is that `getMatType` isn't an additive procedure (like `addJacobian`) so everytime the auxiliary element contribution is computed, it overwrites the element contribution
- The fix to this was to always split out the auxelem matrix contribution and then scale (rather than conditionally doing so)
- I also added a new modalProblem integration test case to catch this in the future

### DIFF
--- a/src/TACSAssembler.cpp
+++ b/src/TACSAssembler.cpp
@@ -4442,7 +4442,6 @@ void TACSAssembler::assembleMatType(ElementMatrixType matType, TACSMat *A,
 
       // Add the contribution from any auxiliary elements,  they need to be
       // scaled first
-      memset(auxElemMat, 0, maxNVar * maxNVar * sizeof(TacsScalar));
       while (aux_count < naux && aux[aux_count].num == i) {
         aux[aux_count].elem->getMatType(matType, i, time, elemXpts, vars,
                                         auxElemMat);
--- a/src/TACSAssembler.cpp
+++ b/src/TACSAssembler.cpp
@@ -4200,23 +4200,21 @@ void TACSAssembler::assembleRes(TACSBVec *residual, const TacsScalar lambda) {
       // Add the residual from any auxiliary elements, if the load factor is 1
       // they can be added straight to the elemRes, otherwise they need to be
       // scaled first
-      if (aux_count < naux) {
-        if (!scaleAux) {
-          while (aux_count < naux && aux[aux_count].num == i) {
-            aux[aux_count].elem->addResidual(i, time, elemXpts, vars, dvars,
-                                             ddvars, elemRes);
-            aux_count++;
-          }
-        } else {
-          memset(auxElemRes, 0, maxNVar * sizeof(TacsScalar));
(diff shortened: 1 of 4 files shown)


src/TACSAssembler.cpp
```diff
@@ -4334,11 +4336,13 @@ void TACSAssembler::assembleJacobian(TacsScalar alpha, TacsScalar beta,
 
       // Add the contribution to the residual and the Jacobian from the
       // auxiliary elements - if any, this is scaled by the loadFactor lambda
-      while (aux_count < naux && aux[aux_count].num == i) {
-        aux[aux_count].elem->addJacobian(i, time, alpha * lambda, beta * lambda,
-                                         gamma * lambda, elemXpts, vars, dvars,
-                                         ddvars, elemRes, elemMat);
-        aux_count++;
+      if (aux_count < naux) {
```
A-CGray, line 4339: Are these statements strictly necessary? I thought C++ uses lazy evaluation of logic statements, so if `aux_count < naux` is false then `aux[aux_count].num == i` won't be evaluated and thus there's no risk of indexing past the bounds of `aux`

timryanb: I've run into segfaults caused by this in the past, see this PR conversation [here](https://github.com/smdogroup/tacs/pull/36#discussion_r723684474). I'm not sure if this is a behavior that has changed in newer C++ versions

timryanb: @A-CGray , you seem to be right about short-circuit evaluation in C++. I added those guards more to be consistent with the other locations in the code where the logic was already being used. The code seems to run w/o issues either way. I'm fine  removing them again if needed.

A-CGray: OK, up to you, both ways should work

tacs/problems/modal.py
```diff
@@ -372,6 +370,8 @@ def _updateAssemblerVars(self):
 
         self.assembler.setDesignVars(self.x)
         self.assembler.setNodes(self.Xpts)
+        # Make previous auxiliary loads are removed
+        self.assembler.setAuxElements(None)
```
A-CGray, line 374: What's the reason for this addition?

timryanb: Other TACS problems (StaticProblem/TranientProblem) may have set an AuxElem loads in a previous call. This line makes sure that AuxElems is empty at the begining of each ModalProblem analysis, since we have no use for them

A-CGray (approved):

PR #173, opened 2023-01-30
Bump min required mpi4py to 3.1.1

A-CGray:
mpi4py 3.1.0 has a typo in it's python dependencies definition that causes errors during pip install (see [here](https://github.com/mpi4py/mpi4py/releases/tag/3.1.1) and [here](https://github.com/mpi4py/mpi4py/commit/e77486bea7a7520d09cea48cd03a98d583a85f60)). Due to a new [major release](https://pypi.org/project/setuptools/#history) of setuptools (best guess), this just started causing build errors on our docker images (see [here](https://github.com/mdolab/docker/actions/runs/4040746729/jobs/6946677857#step:10:9815))

This PR therefore bumps the minimum required version of mpi4py to 3.1.1 where the error is fixed.

Also includes the OpenMDAO version guard from #174 

Bumping TACS version to 3.2.0

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,4 +1,4 @@
 #pyproject.toml
 [build-system]
 # Minimum requirements for the build system to execute.
-requires = ['setuptools>=45.0', 'wheel', 'cython>=0.29', 'oldest-supported-numpy', 'mpi4py==3.1.0']
+requires = ['setuptools>=45.0', 'wheel', 'cython>=0.29', 'oldest-supported-numpy', 'mpi4py==3.1.1']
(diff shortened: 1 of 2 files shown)


timryanb:
3.1.1 as a minimum seems fine to me. I don't see anything in the release notes for (3.1.1+) that would affect TACS and want to avoid forcing users to upgrade unless there's a specific reason

timryanb (approved):

timryanb:
I'm going to remove the version bump and merge this PR as-is for now to fix the broken installation. I'll make a new separate PR with the version bump (PR #177)

PR #176, opened 2023-01-31
modify tacs wrapper for thermal

joanibal:
This PDR is to add the capability to do steady conduction analysis to the MPhys TACS wrapper


joanibal:
I changed it to a draft because now I need to verify that it still works after the merge

joanibal:
@timryanb would you mind reviewing this? I can't add reviewers

PR #185, opened 2023-02-18
Add methods for retreiving number of iterations and residual norm from KSM linear solvers

### LINKED ISSUE #4: KSM Solver choice and iteration count
We currently have an implementation of a nonlinear solver which works with the old version of TACS that I am attempting to transfer over to this newer version. In our solver, we check the number of iterations taken during each linear solve in order to decide whether to update the preconditioner. This was done using a `getIterCount` method inside the GCROT class, this method was not implemented in the GMRES or PCG classes although as far as I can see there is an iteration counting variable that could be output.

The problem I am running into now is that although GCROT, GMRES and PCG are all still defined in KSM.cpp, the python interface is currently hardcoded to return a GMRES object when asked for a KSM:
```python
cdef class KSM:
    def __cinit__(self, Mat mat, Pc pc, int m,
                  int nrestart=1, int isFlexible=0):
        '''
        Create a GMRES object for solving a linear system with or
        without a preconditioner.

        This automatically allocates the requried Krylov subspace on
        initialization.

        input:
        mat:        the matrix operator
        pc:         the preconditioner
        m:          the size of the Krylov subspace
        nrestart:   the number of restarts before we give up
        isFlexible: is the preconditioner actually flexible? If so use FGMRES
        '''

        self.ptr = new GMRES(mat.ptr, pc.ptr, m, nrestart, isFlexible)
        self.ptr.incref()
        return

    def __dealloc__(self):
        if self.ptr:
            self.ptr.decref()
        return
```
I therefore have 2 questions:
- Are there some bugs with GCROT and PCG that are behind the reason for making only GMRES available through the python interface?
- If not, do you have anything against me adding the functionality to choose different KSM's through the interface and also adding a `getIterCount` method to each one?

Thanks

Ali

A-CGray:
In my nonlinear solver, I wanted to be able to check the final residual and the number of iterations from each linear solve (e.g to decide whether the preconditioner needs to be updated), this PR adds this functionality for all solvers that inherit from `TACSKsm` 

PR Summary:
- Adds protected `iterCount` and `resNorm` attributes to the abstract `TACSKsm` class, `iterCount` is the total number of iterations taken for the last linear solve and `resNorm` is the residual norm from the end of the last linear solve.
- Adds default constructor to `TACSKsm` that initialises `iterCount` and `resNorm` to 0
- Adds getter methods for `iterCount` and `resNorm` to `TACSKsm` and their python interfaces
- Adds code to update  `iterCount` and `resNorm` to the end of the `solve` methods of the `PCG`, `GMRES` and `GCROT` classes that inherit from `TACSKsm` 
- Adds basic convergence tests that rely on `iterCount` and `resNorm` to `StaticTestCase.test_solve`

Closes #4

### DIFF
--- a/src/bpmat/KSM.cpp
+++ b/src/bpmat/KSM.cpp
@@ -456,6 +456,7 @@ int PCG::solve(TACSVec *b, TACSVec *x, int zero_guess) {
 
     if (count == 0) {
       rhs_norm = R->norm();
+      resNorm = rhs_norm;
     }
 
     if (monitor && count == 0) {
@@ -480,20 +481,19 @@ int PCG::solve(TACSVec *b, TACSVec *x, int zero_guess) {
         P->axpby(1.0, beta, Z);                    // P' = Z' + beta*P
         iterCount++;
 
-        TacsScalar norm = R->norm();
+        resNorm = R->norm();
 
         if (monitor) {
-          monitor->printResidual(i + 1, norm);
+          monitor->printResidual(i + 1, resNorm);
         }
 
-        if (TacsRealPart(norm) < atol ||
-            TacsRealPart(norm) < rtol * TacsRealPart(rhs_norm)) {
+        if (TacsRealPart(resNorm) < atol ||
+            TacsRealPart(resNorm) < rtol * TacsRealPart(rhs_norm)) {
           solve_flag = 1;
           break;
         }
       }
     }
-    resNorm = norm
 
     if (solve_flag) {
       break;
@@ -931,7 +931,6 @@ int GMRES::solve(TACSVec *b, TACSVec *x, int zero_guess) {
       pc->applyFactor(work, W[0]);
       x->axpy(1.0, W[0]);
     }
-
 
     iterCount += niters;
     resNorm = res[niters];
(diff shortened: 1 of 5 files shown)


timryanb (approved):

src/bpmat/KSM.cpp
```diff
@@ -928,6 +932,9 @@ int GMRES::solve(TACSVec *b, TACSVec *x, int zero_guess) {
       x->axpy(1.0, W[0]);
     }
 
+    iterCount += niters;
+    resNorm = res[niters];
```
gjkennedy, line 936: You should take the absolute value of this. "res" really stores the solution here since the Givens rotations are applied to it. The norm should always be positive.

src/bpmat/KSM.cpp
```diff
@@ -1395,6 +1404,8 @@ int GCROT::solve(TACSVec *b, TACSVec *x, int zero_guess) {
     // mat->mult( x, R );      // R = A*x
     // R->axpy( -1.0, b );     // R = A*x - b
     // R->scale( -1.0 );
+    iterCount += niters;
+    resNorm = R->norm();
```
gjkennedy, line 1408: No need to compute this norm outside the solve flag right?

gjkennedy (commented):
A few minor corrections

src/bpmat/KSM.cpp
A-CGray: Should I move these lines above this solution computation block to get the true residual then?

src/bpmat/KSM.cpp
A-CGray: Fixed

A-CGray:
@gjkennedy are you happy with merging this now?

gjkennedy (approved):

PR #187, opened 2023-02-22
Add `__init__` method to `BaseUI` for setting up pyTACS options/comm

A-CGray:
Previously, all pyTACS classes had the same boilerplate code in their `__init__` methods for setting up their MPI communicators and options dictionaries. This PR moves that code into the `__init__` method of the `BaseUI` class which is inherited by all pyTACS classes, reducing repeated code.

### DIFF
--- a/tacs/problems/base.py
+++ b/tacs/problems/base.py
@@ -19,7 +19,9 @@ class TACSProblem(BaseUI):
     Base class for TACS problem types. Contains methods common to all TACS problems.
     """
 
-    def __init__(self, assembler, comm, outputViewer=None, meshLoader=None):
+    def __init__(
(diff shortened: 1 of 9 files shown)


A-CGray:
Thanks for those fixes @timryanb. Out of interest, is there a specific reason you prefer calling `TACSProblem.__init__` over `super().__init__`?

timryanb:
> Thanks for those fixes @timryanb. Out of interest, is there a specific reason you prefer calling `TACSProblem.__init__` over `super().__init__`?

I think the second is a little more explicit than the super approach. Makes it clearer what class method is being called. 

It's also consistent with other locations in the code, like here: https://github.com/smdogroup/tacs/blob/master/tacs/problems/static.py#L401

timryanb (approved):

PR #189, opened 2023-03-03
Truncating PBAR shear correction value

timryanb:
pyNastran defaults the shear correction values to 1e8 (meant to approximate infinity to recover EB beam behavior) for PBAR cards if the values are not specified in the bdf file. This high value can lead to poor scaling in the stiffness matrix, which can lead to numerical issues in the solution. To fix this I'm adding a truncation step in `_elemCallBackFromBDF` that will use a smaller value of 1e3 instead.


A-CGray:
> pyNastran defaults the shear correction values to 1e8 (meant to approximate infinity to recover EB beam behavior) for PBAR cards if the values are specified in the bdf file.

Just checking my understanding, this should say "if the values are **not** specified in the bdf file" right? So if the user specifies correction factors in their BDF those will be correctly passed on to TACS?

gjkennedy (approved):

PR #194, opened 2023-03-31
Time spectral finite-element implementation

gjkennedy:
This code implements a time-spectral approach for integrating the equations of motion in time. This is only for first-order systems in time (such as thermal problems) and could be extended to second-order systems or periodic systems.

### DIFF
--- a/src/bpmat/BCSRMat.h
+++ b/src/bpmat/BCSRMat.h
@@ -56,6 +56,9 @@ class BCSRMat : public TACSObject {
   // Free the matrix
   ~BCSRMat();
 
+  // Create a duplicate of the matrix without copying values
+  BCSRMat *createDuplicate();
+
   // Access data
   // -----------
   MPI_Comm getMPIComm();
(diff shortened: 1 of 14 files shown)


src/bpmat/BCSRMat.h
```diff
@@ -56,6 +56,9 @@ class BCSRMat : public TACSObject {
   // Free the matrix
   ~BCSRMat();
 
+  // Create a duplicate of the matrix without copying values
+  BCSRMat *createDuplicate();
+
```
komahanb, line 61: @gjkennedy using Fortran terminology: a mold() function would allocate but doesn't copy values

gjkennedy: I think duplicate is a nice description for this. This function will be used infrequently and typically not in instances when you actually want to copy the values of the matrix.

timryanb (approved):
Nice, we'll have to add a problem wrapper for this in pytacs at some point

PR #209, opened 2023-05-12
Add method to get output parameters from ESP/CAPS through tacsAIM

bburke38:


sean-engelstad (approved):
Looks good, definitely is important to measure output parameters from ESP/CAPS which can be conditional functions of ESP/CAPS design parameters.

gjkennedy (approved):

PR #212, opened 2023-05-19
Add bladeStiffenedShell constitutive model

A-CGray:
This PR adds a new `TACSBladeStiffenedShellConstitutive` class, a new implementation of the old `bladeFSDT` class. It also adds a panel length constraint class to the pyTACS interface.

## Constitutive model
Compared to the old `bladeFSDT` class, the new model contains a few improvements:

1. The panel and stiffener laminates can now be made up of an arbitrary number of ply angles as opposed to being hardcoded to [0, -45, 45, 90] deg laminates. This means you can model more complex laminates and makes it simpler to model a shell made from an isotropic material as you no longer need to supply 4 copies of the same isotropic ply when creating the constitutive object.
2. The buckling failure envelope is modified in a similar manner to https://github.com/smdogroup/tacs/pull/203 so that it can be used to ensure a specified buckling margin
3. I fixed a bug (I think) in the critical buckling loads calculation where the $D_{12}$ entry of the bending stiffness was being used instead of $D_{22}$ to compute the $D_2$ value in Stroud and Agranoff's equations.
4. Almost all of the calculations in the model have been rewritten in a way that computes contributions from the panel and stiffener as if they are distinct shell and beam constitutive models. My hope is that in the future, with a little more work, we could modify this model to create a stiffened shell model using any combination of existing TACS shell and beam constitutive models. 

For example, this is how the stiffness matrix for the stiffened shell is computed:
```cpp
// Compute the stiffness matrix
void TACSBladeStiffenedShellConstitutive::computeStiffness(TacsScalar C[]) {
  TacsScalar* A = &C[0];
  TacsScalar* B = &C[6];
  TacsScalar* D = &C[12];
  TacsScalar* As = &C[18];
  TacsScalar* drill = &C[21];

  // --- Zero out the C matrix ---
  memset(C, 0, this->NUM_TANGENT_STIFFNESS_ENTRIES * sizeof(TacsScalar));

  // Add the panel's stiffness contributions
  this->computePanelStiffness(C);

  // Compute the stiffener's beam stiffness matrix ,then transform it to a shell
  // stiffness matrix and add it
  TacsScalar Cstiff[TACSBeamConstitutive::NUM_TANGENT_STIFFNESS_ENTRIES];
  this->computeStiffenerStiffness(Cstiff);
  this->addStiffenerStiffness(Cstiff, C);
}
```
This code does not make any assumptions about the form of the panel or stiffener stiffness matrices and could easily be modified so that `computePanelStiffness` is actually a call to another shell constitutive model, and `computeStiffenerStiffness` is a call to a beam constitutive model.

## Panel length constraints
I also implemented a panel length constraint class that is necessary to ensure the panel length values used in the buckling failure calculations are consistent with the true geometry of each TACS component. I used [numba](https://numba.pydata.org/) to speed up the panel length computation and its derivatives which means we would need to add Numba as a dependency of TACS. As far as I can tell this shouldn't be too much of a stability issue, numba has a pretty wide user-base, and is officially supported by organisations like Anaconda, Intel, and NVIDIA.

## ToDo before merging:

- [x] Merge #203 
- [x] Merge #216 
- [x] Add numba dependency
- [x] Fix Tsai-Wu derivative errors
- [x] Add integration test for panel-length constraints
- [x] Add integration test for BladeStiffenedShell
- [x] Add BladeStiffenedShell example case?


tacs/pytacs.py
```diff
@@ -1948,6 +1948,36 @@ def createDVConstraint(self, name, options=None):
...
+        ----------
+        name : str
+            Name to assign constraint.
+        options : dict
+            Class-specific options to pass to DVConstraint instance (case-insensitive).
+
+        Returns
+        ----------
+        constraint : DVConstraint
+            DVConstraint object used for calculating constraints.
```
timryanb, line 1966: Update docstring here. should be `tacs.constraints.PanelLengthConstraint`

timryanb (commented):
We'll have to add a new integration test that tests the panel length constraint

timryanb:
Any idea whats going on with the constitutive test failures?

timryanb:
Out of curiosity how long does it take to compute the panel length constraints for your CRM case?

timryanb:
Can you add a conda dependency line for numba [here](https://github.com/smdogroup/tacs/blob/master/conda/meta.yaml#L66-L80)?

timryanb:
Other than that. This looks good so far. Thanks for the work @A-CGray

A-CGray:
@timryanb on further inspection, there was no issue with the Tsai-Wu sensitivities but there was an issue with the Von Mises (not related to any changes in this PR) that I fixed. There were some issues with the finite difference tests for the blade stiffened model where the relative tolerance cannot be met because the absolute values of the derivatives are very small. To get around this I implemented a new type of pass/fail criteria for the unit tests based on numpy's [assert_allclose](https://numpy.org/doc/stable/reference/generated/numpy.testing.assert_allclose.html). It basically acts like a relative tolerance when the values being compared are large but more like an absolute tolerance when the compared values are very small which fixes the issue I was facing.

A-CGray:
I stand corrected, there were a couple of uninitialized values causing problems only in the real tests when compiled in debug mode. I've fixed those and with the new `assert_allclose` check I was able to get all of the constitutive model tests passing with tighter tolerances

A-CGray:
@timryanb I think once #216 is merged, the tests here should pass and this PR will be ready to go

tests/constitutive_tests/test_iso_tube_beam_constitutive.py
```diff
@@ -9,15 +9,15 @@ class ConstitutiveTest(unittest.TestCase):
...
+            self.rtol = 1e-12
         else:
-            self.dh = 1e-6
-            self.rtol = 1e-2
-            self.atol = 1e-1
+            self.dh = 1e-8
+            self.rtol = 1e-6
         self.dtype = TACS.dtype
 
+        self.atol = np.clip(1e-5 * self.rtol, 1e-8, 1e-14)
```
timryanb, line 19: Will we need these clip statements anymore once PR #216 gets merged @A-CGray ?

A-CGray: Kind of, they are added in that PR

timryanb:
@A-CGray I had to rename the benchmark file to get `testflo -b` to work since we already had a file called `benchmark_analysis.py` and testflo didn't like that.

examples/mach_tutorial_wing/analysis.py
```diff
@@ -0,0 +1,397 @@
...
+        lower=-heightAdjCon,
+        upper=heightAdjCon,
+        dvIndex=3,
+    )
+    adjCon.addConstraint(
+        conName=group + "_stiffenerPitchAdj",
+        compIDs=compIDs[group],
+        lower=-pitchAdjCon,
+        upper=pitchAdjCon,
+        dvIndex=2,
```
timryanb, line 329: Is this a typo @A-CGray ? Both the pitch and panel thickness seem to be tagged as dvIndex 2 here

gjkennedy (approved):

Issue #217, opened 2023-06-01
Break up MPhys wrapper


timryanb:
The `mphys_tacs.py` file under `tacs/mphys` is getting a bit long and probably could be better named.
For better readability, we should break the file down into smaller submodules (builder.py, solver.py, functions.py?) with each component living in its own file. We can still keep the import statement at the user level the same using the `__init__.py` (i.e. `from tacs.mphys import TacsBuilder`)

PR #220, opened 2023-06-06
Breaking mphys script into multiple

timryanb:
Addresses Issue #217


tacs/mphys/solver.py
```diff
@@ -0,0 +1,196 @@
...
+                    )
+
+                if "tacs_dvs" in d_inputs:
+                    self.sp.addAdjointResProducts(
+                        [d_residuals[self.states_name]],
+                        [d_inputs["tacs_dvs"]],
+                        scale=1.0,
+                    )
+
+    def _design_vector_changed(self, x):
```
A-CGray, line 185: Is this still needed or has it been replaced by `_need_update`?

A-CGray (changes requested):

tacs/mphys/solver.py
timryanb: Good point, it never seems to be used. Should be removed now

A-CGray (approved):

PR #216, opened 2023-06-01
Unit testing tolerance improvements + Von Mises sensitivity fix

A-CGray:
In unit tests we currently use an essentially infinite absolute tolerance and a very large relative tolerance. This is done because it is very hard to specify a single absolute tolerance that is valid for test values that span multiple orders of magnitude, and relative errors can be very high even for accurate derivatives when the test values are very small. 

This PR fixes this issue by implementing a new kind of check between test and reference values. In doing so, the constitutive unit tests now pass with tolerances 1000x tighter than before.

The check now performed is identical to that used by numpy's [assert_allclose method](https://numpy.org/doc/stable/reference/generated/numpy.testing.assert_allclose.html):

`fail = abs( test - ref ) > atol + rtol * abs(ref)`

Where `ref` is the finite-difference or complex-step value and `test` is the analytic value

When `abs(ref) >> atol/rtol` this check acts like a relative tolerance check, but for `abs(ref) << atol/rtol` it acts like an absolute tolerance check, thus avoiding the issues associated with relative tolerances of very small values.

## Changes:

- Add a `TacsAssertAllClose` method 
- Switch to using `TacsAssertAllClose` for all constitutive model unit tests
- Reduce relative tolerance from 1e-1 to 1e-4
- Reduce absolute tolerance from 1e99 to `1e-5*self.rtol` clipped to be within 1e-8 - 1e-14
- Reduce finite difference step size from 1e-6 to 1e-8
- Add absolute error to output of `TacsPrintErrorComponents`
- Reduce complex-step step size from 1e-50 to 1e-200
- Reduce magnitude of strains used in constitutive tests from 1 to 1e-3
- Fix bug in Von Mises strain sensitivity that was previously missed when testing with huge strains

PR #221, opened 2023-06-09
Adding smeared composite constitutive class

timryanb:
Adding smeared composite shell constitutive class where layup is defined based on ply fractions rather than stacking sequence. Useful for optimization where we want to ignore the discrete nature of laminate design.


src/constitutive/TACSSmearedCompositeShellConstitutive.h
```diff
@@ -0,0 +1,145 @@
...
+
+  // Evaluate the specific heat
+  TacsScalar evalSpecificHeat(int elemIndex, const double pt[],
+                              const TacsScalar X[]);
+
+  // Evaluate the stress
+  void evalStress(int elemIndex, const double pt[], const TacsScalar X[],
+                  const TacsScalar strain[], TacsScalar stress[]);
+
+  // Add the contribution
```
A-CGray, line 78: Incomplete docstring

src/constitutive/TACSSmearedCompositeShellConstitutive.cpp
```diff
@@ -0,0 +1,600 @@
...
+  }
+
+  TacsScalar ks_val =
+      ksAggregationSensProduct(fvals, nvals, nvars, ks_weight, dfvals, sens);
+
+  for (int i = 0; i < nvals; i++) {
+    delete[] dfvals[i];
+  }
+
+  delete[] fvals;
```
A-CGray, line 478: It might not matter since these partials are generally much cheaper than the adjoint solve, but I think allocating and deleting memory like this inside functions that are going to be called inside "hot" loops is not good for performance. In the `BladeStiffenedShell` class I got around this by making these arrays attributes of the class that get created/destroyed at the same time as the constitutive object

src/constitutive/TACSSmearedCompositeShellConstitutive.cpp
```diff
@@ -0,0 +1,600 @@
...
+      // plate top stress
+      getLaminaStrain(e, tt, lamStrain);
+      TacsScalar fval_top =
+          ply_props[i]->failureStrainSens(ply_angles[i], lamStrain, phi);
+      dfdx[index] += 0.5 * scale * dks_vals[2 * i + 1] *
+                     (phi[0] * e[3] + phi[1] * e[4] + phi[2] * e[5]);
+    }
+
+    delete[] fvals;
+    delete[] dks_vals;
```
A-CGray, line 519: Same comment as above with regard to allocating and destroying memory in functions

tests/integration_tests/test_shell_comp_smeared.py
```diff
@@ -0,0 +1,190 @@
...
+
+        # Overwrite default check values
+        if self.dtype == complex:
+            self.rtol = 1e-6
+            self.atol = 1e-6
+            self.dh = 1e-50
+        else:
+            self.rtol = 2e-1
+            self.atol = 1e-3
+            self.dh = 1e-6
```
A-CGray, line 56: Maybe update these to bring them in line with the changes made in #216

tests/integration_tests/test_shell_comp_write_bdf.py
```diff
@@ -44,8 +44,8 @@ def setup_tacs_problems(self, comm):
 
         # Overwrite default check values
         if self.dtype == complex:
-            self.rtol = 1e-8
-            self.atol = 1e-8
+            self.rtol = 1e-7
+            self.atol = 1e-7
             self.dh = 1e-50
         else:
             self.rtol = 2e-1
```
A-CGray, line 51: Can we make these tolerances tighter?

A-CGray (changes requested):

tests/integration_tests/test_shell_comp_write_bdf.py
timryanb: Unfortunately, not. For whatever reason, the accuracy tops out a little above 1e-8. I think we should take this accuracy for now

A-CGray: Sorry, this comment got applied to the wrong lines, I meant the real test values

timryanb: Ah gotcha, yes we I can probably drop that a bit

src/constitutive/TACSSmearedCompositeShellConstitutive.h
timryanb: @A-CGray, could you clarify this a bit? What are you expecting here?

timryanb: Nvm, I assume you mean adding docstring parameters of the C++ side? 
I can do this, but I'm noticing there's a number of classes that are missing these docstrings as well ([TACSIsoShellConstitutive.h](https://github.com/smdogroup/tacs/blob/master/src/constitutive/TACSIsoShellConstitutive.h), [TACSIsoTubeBeamConstitutive.h](https://github.com/smdogroup/tacs/blob/master/src/constitutive/TACSIsoTubeBeamConstitutive.h), etc).

We should go back (in a separate PR) and make this standard consistent across all C++ classes

A-CGray: I just meant that this docstring ("Add the contribution") doesn't fully describe what the function does. You're right that we should probably go through and standardize how we write docstrings in the header and source files respectively though

src/constitutive/TACSSmearedCompositeShellConstitutive.cpp
timryanb: I've moved all of these arrays to the construction/destruction method for the class

src/constitutive/TACSSmearedCompositeShellConstitutive.cpp
timryanb: See above

src/constitutive/TACSSmearedCompositeShellConstitutive.h
timryanb: I fixed the truncated docstring here. We can leave documenting the parameters consistently across all C++ classes for another PR

A-CGray (approved):

gjkennedy (approved):

PR #225, opened 2023-06-30
Add helper scripts for creating and deleting large numbers of tecplot files

A-CGray:
I have a bunch of helper scripts that I use for processing the large number of f5 files that get output from my optimisations. This PR would add them to the repo so other people can use them too.

## Summary of scripts:

- `f5convert_all`: Runs `f5totec` in parallel on all .f5 files in the current folder for which there is not an existing tecplot file (or there is a tecplot file but it's older than the f5 file) 
- `f5convert_subdirs`: Finds all directories from the current directory down that contain .f5 files and runs f5convert_all in each one.
- `f5clean_all`: Deletes all tecplot files in the current directory for which there is a corresponding f5 files (useful for freeing up disk space)
- `f5clean_subdirs`: Finds all directories from the current directory down that contain .f5 files and runs f5clean_all in each one.


timryanb:
Thanks @A-CGray, I think these features would be a good addition. 
Is there anyway we could implement them under f5totec as command line options?
We could also potentially add this to f5tovtk

A-CGray:
> Thanks @A-CGray, I think these features would be a good addition. Is there anyway we could implement them under f5totec as command line options? We could also potentially add this to f5tovtk

I'm not sure how easy it'd be to add these capabilities to the f5totec executable as that would require re-implementing all the filesystem operations in C.

However, it should be pretty easy to add a command line-option to the scripts to switch between using f5totec and f5tovtk. I could probably also make the subdirectory traversal a command-line option, so we'd just end up with two scripts `f5convert` and `f5clean`

timryanb:
> > Thanks @A-CGray, I think these features would be a good addition. Is there anyway we could implement them under f5totec as command line options? We could also potentially add this to f5tovtk
> 
> I'm not sure how easy it'd be to add these capabilities to the f5totec executable as that would require re-implementing all the filesystem operations in C.
> 
> However, it should be pretty easy to add a command line-option to the scripts to switch between using f5totec and f5tovtk. I could probably also make the subdirectory traversal a command-line option, so we'd just end up with two scripts `f5convert` and `f5clean`

Alright, that works for me

timryanb (approved):
Let's give @gjkennedy a chance to review

gjkennedy (approved):

PR #239, opened 2023-08-03
Conda numba build fix

timryanb:
- anaconda sometimes pulls old versions of numba which break the install
- Adding version pin for numba to conda recipe

### DIFF
--- a/conda/meta.yaml
+++ b/conda/meta.yaml
@@ -77,6 +77,34 @@ requirements:
     - pynastran
     - numba >=0.55.2
 
+test:
+  requires:
+    - testflo
+
+  source_files:
+    - tests
+
+  imports:
+    - tacs
+    - tacs.pytacs
+    - tacs.problems
+    - tacs.constraints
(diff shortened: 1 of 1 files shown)


conda/meta.yaml
```diff
@@ -75,7 +75,7 @@ requirements:
     - metis ==5.1.0
     - mpi4py
     - pynastran
-    - numba
+    - numba >=0.55.2
```
A-CGray, line 78: Should we add the same minimum version to the setup.py requirements?

timryanb: I dont think so. This appeared to be an issue specifically with the conda packages dependencies for older versions of numba that didn't have a version upperbound for newer numpy versions. This could lead to a situation where an old version of numba could be installed with a new (incompatible) version of numpy.

A-CGray: ok, if it's a very conda-specific issue then I guess we can go with this as is

PR #248, opened 2023-09-14
Adding method to set finite difference step size used in element classes

### LINKED ISSUE #246: Expose finite difference step size as an option in buckling problem
Would it be possible to expose the finite difference step size as a keyword argument when setting up a buckling problem? I think this could help resolve the issue of discrepancies between load factors/buckling modes between TACS and commercial codes.

timryanb:
- Adding method to set the finite difference step size for TACSElements
- The finite difference step size used to estimate the geometric stiffness in TACShellElement is now tied to the TACSElement fd step size
- The step size all elements in a model can be modified by calling this class method at the begining of a script
```
from tacs.TACS import Element
Element.setFiniteDifferenceStepSize(1e-6)
```
- The change in step size in the geometric stiffness routine has caused a few of our more sensitive buckling sensitivity tests to fail
- Still have to figure out a way to work around this
- Closes #246

### DIFF
--- a/src/elements/TACSElement.cpp
+++ b/src/elements/TACSElement.cpp
@@ -28,9 +28,23 @@
 int TACSElement::fdOrder = 2;
 
 /*
+  Default finite difference step length
+*/
+#ifdef TACS_USE_COMPLEX
+double TACSElement::dh = 1e-30;
+#else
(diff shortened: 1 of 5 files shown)


timryanb:
Closing now that PR #304 is merged

Issue #249, opened 2023-09-15
Orthotropic heat conduction in shell elements


Pranav4495:
Hi,

I believe the current version can handle heat conduction problems with isotropic thermal conductivities (kappa). Specifying kappa1 and kappa2 instead of kappa will resort to the predefined value of kappa in the code. Is there a way to use kappa1 and kappa2 in 2D and kappa1, kappa2 and kappa3 in 3D conduction problems?

Thanks,
Pranav

timryanb:
It looks like there was an oversight in how orthotropic properties are handled in the `MaterialProperties` class (see [here](https://github.com/smdogroup/tacs/blob/master/tacs/constitutive.pyx#L316-L324)). It looks like the code uses decides whether or not the provided properties are orthotropic or not based on if the 1-1 Young's modulus "E1" is included in kwargs. 
Obviously, you shouldn't need to provide this parameter for a purely thermal analysis, so in the future we can fix this to make it more general. 

 In the meantime, if you add E1 to your kwargs when instantiating your `MaterialProperties` (even if you don't need it), you should get the orthotropic properties you're expecting.

```
prop = MaterialProperties(E1=1.0, kappa1=1e-6, kappa2=2e-6, ....)
```

Pranav4495:
That clarifies it. Thanks for the quick response. I will try and let you know how it goes.

Pranav4495:
I tried the 'thermal_plate' example by specifying E1 and using kappa1=kappa2=kappa. Ideally, it should yield the same results as specifying kappa=kappa, but it seems there are convergence issues with the solver, probably due to ill-conditioning. Please see the attached image. I have tried a different analysis, giving similar issues.

Additionally, I have a few questions/comments on the problem setup.

1. Is there a way to define lamina ply angles and material properties as design variables? If so, will the sensitivities w.r.t these variables be calculated internally?
2. I am trying to use the buckling solver with sigma=0 (shift-invert mode), giving all zero eigenvalues for the first 10 modes, whereas specifying a non-zero small positive sigma yields eigenvalues that are very different from what I get using Nastran. The static analysis part of the same problem works fine, so I am wondering if there is anything I am missing in the problem buckling problem definition. 

<img width="648" alt="ht_ortho_illcondition" src="https://github.com/smdogroup/tacs/assets/70219987/88354b2b-6402-4466-9ff5-910259086314">

timryanb:
[...]

PR #252, opened 2023-09-20
Fixing kappa initialization bug in MaterialProperties

timryanb:
- kappa1, kappa2, and kappa3 are never actually set for orthotropic materials during initialization
- Addresses #249

### DIFF
--- a/src/constitutive/TACSMaterialProperties.cpp
+++ b/src/constitutive/TACSMaterialProperties.cpp
@@ -113,6 +113,10 @@ TACSMaterialProperties::TACSMaterialProperties(
   alpha1 = _alpha1;
   alpha2 = _alpha2;
   alpha3 = _alpha3;
+
+  kappa1 = _kappa1;
+  kappa2 = _kappa2;
+  kappa3 = _kappa3;
 }
 
 // Get the material type


tacs/constitutive.pyx
```diff
@@ -215,6 +215,14 @@ cdef class MaterialProperties:
         cdef TacsScalar kappa2 = 230.0
         cdef TacsScalar kappa3 = 230.0
 
+        ISOTROPIC_KEYS = ["rho", "E", "nu", "G", "ys", "alpha", "specific_heat", "kappa"]
```
Pranav4495, line 218: Defining 'specific_heat' for orthotropic material would result in 'is_orthotropic=False' due to it being included in ISOTROPIC_KEYS.

timryanb: That's only the case if _only_ `specific_heat` is provided. If any keys are provided that aren't a part of this list, then it defaults to orthotropic (see [here](https://github.com/timryanb/tacs/blob/init_bug/tacs/constitutive.pyx#L221)).

For instance:
```
mat = MaterialProperties(specific_heat=0.5) # isotropic
mat = MaterialProperties(specific_heat=0.5, kappa1=0.2) # orthotropic
```

Pranav4495: My bad. I interpreted the if statement incorrectly. Thanks for clarifying.

gjkennedy (approved):

PR #254, opened 2023-09-28
add orthotropic material properties for use in funtofem+caps2tacs

sean-engelstad:
* I've added orthotropic materials to the caps2tacs materials module. 
* Marshall and I added MAT4, MAT5 thermal cards into the tacsAIM which have material IDs that are 100 + the material IDs for the elastic cards (so they are tied together with that property group).
* We can now specify thermal properties for orthotropic materials in caps2tacs => funtofem (namely kappa1 - kappa3, and specific_heat)
* Question: Can we add the material ID = 100 + mid checks into the default pytacs callback too? This would allow us to specify thermal properties from caps2tacs into pure TACS as well. @bburke38 and I think this would be a good idea, but what do you think @timryanb

### DIFF
--- a/tacs/caps2tacs/materials.py
+++ b/tacs/caps2tacs/materials.py
@@ -8,16 +8,28 @@ class Material:
         self,
         name: str,
         material_type: str,
-        young_modulus: float,
-        poisson_ratio: float,
-        density: float,
-        tension_allow: float,
-        compression_allow: float = None,
-        shear_allow: float = None,
-        yield_allow: float = None,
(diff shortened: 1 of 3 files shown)


timryanb:
I'd prefer to leave the matID + 100 procedure outside of pyTACS. I want to preserve NASTRAN compatibility there. I'd be concerned about cases with large numbers of material cards in the BDF file where two unrelated mat cards unintentionally get read together.

timryanb (approved):

PR #258, opened 2023-10-05
Adding warning to `MaterialProperties` for invalid keyword args

timryanb:
- Since PR #252 , invalid input args being passed to MaterialProperties change the default material type to orthotropic
- This can have unexpected behavior for users who accidentally misspell valid isotropic parameters 
- To make the issue clearer the following changes have been made:
    1. Passing an invalid/misspelled argument raises an explicit warning
    2. Invalid arguments are ignored when deciding if the properties are orthtropic/isotropic

### DIFF
--- a/tacs/constitutive.pyx
+++ b/tacs/constitutive.pyx
@@ -235,7 +235,7 @@ cdef class MaterialProperties:
               warnings.warn(
                 f"Invalid keyword argument '{input_key}' detected in MaterialProperties. \n"
                 "Keyword argument will be ignored. Acceptable arguments are: \n"
-                f"{ALL_KEYS}",
+                + ", ".join(ALL_KEYS),
                 RuntimeWarning
             )
 
--- a/tacs/constitutive.pyx
+++ b/tacs/constitutive.pyx
@@ -13,6 +13,7 @@
 #distutils: language=c++
 
 import warnings
+import difflib
 
 # For the use of MPI
 from mpi4py.libmpi cimport *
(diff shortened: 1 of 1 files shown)


tacs/constitutive.pyx
```diff
@@ -216,9 +218,29 @@ cdef class MaterialProperties:
...
+                            "S12", "S13", "S23",
+                            "Xt", "Yt", "Xc", "Yc", "S",
+                            "alpha1", "alpha2", "alpha3",
+                            "kappa1", "kappa2", "kappa3"]
+        ALL_KEYS = ISOTROPIC_KEYS + ORTHOTROPIC_KEYS
+
+        # Check for any invalid/misspelled inputs
+        for input_key in kwargs:
+          if input_key not in ALL_KEYS:
+              warnings.warn(
```
A-CGray, line 235: I would prefer this result in an exception than a warning, it's easy to miss a warning amongst a bunch of other output. If a material property I thought I was setting is not getting set then whatever analysis I'm running is pretty much guaranteed to produce a useless result anyway.

timryanb: I don't disagree with this. I do worry that it would be a breaking change (though an easy one to fix) and could lead to a lot of disruption downstream. 

I think if we were going to go the error route, we should just refactor to get rid of kwargs and explicitly define expected arguments in the function handle anyways, which would be a bigger change.

timryanb: @A-CGray I'm fine with leaving it an error

A-CGray (approved):

PR #267, opened 2023-10-30
Add ability to animate shape variables to a gif in caps2tacs

sean-engelstad:
* New routine `tacs_model.animate_shape_variables` can be used to help make gif files of a range of structural shape variables. 
* No forward/adjoint analyses need to be run as that would slow down the animation process. Instead, a `writeDummySensFile` method is included since the tacsAIM postAnalysis expects function and derivative values before creating a new design.
* These features were successfully tested out on the `simple_naca_wing.csm` geometry in `examples/caps_wing/7_animate_shape_variables.py` (a new run script).
* This is useful in debugging the shape variable design spaces in caps2tacs.

### DIFF
--- a/tacs/problems/base.py
+++ b/tacs/problems/base.py
@@ -935,3 +935,46 @@ class TACSProblem(TACSSystem):
                                 hdl.write("1\n")
                                 hdl.write(f"{struct_sens[idx].real}\n")
             return
+
+    def writeDummySensFile(self, tacsAim):
+        """
+        write a blank / dummy SensFile for caps2tacs to enable animating shape variables
+        using GIF animations (without having to run an analysis just building each mesh)
+
+        Parameters
+        ----------
+        tacsAim : tacs.caps2tacs.TacsAIM
+            class which handles the sensitivity file writing for ESP/CAPS shape derivatives
+
+        """
+
+        num_funcs = 1
+        assert tacsAim is not None
+        num_struct_dvs = len(tacsAim.thickness_variables)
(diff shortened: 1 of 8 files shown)


tacs/problems/base.py
```diff
@@ -935,3 +935,46 @@ class which handles the sensitivity file writing for ESP/CAPS shape derivatives
...
+        num_funcs = 1
+        assert tacsAim is not None
+        num_struct_dvs = len(tacsAim.thickness_variables)
+        num_nodes = self.meshLoader.bdfInfo.nnodes
+
+        if self.comm.rank == 0:
+            # open the sens file nastran_CAPS.sens and write coordinate derivatives
+            # and any other struct derivatives to it
+            with open(tacsAim.sens_file_path, "w") as hdl:
+                for func_name in ["mass"]:
```
timryanb, line 960: This looks like a lot of repeated code from the above `writeSensFile` function. Can you just re-use that function? Maybe allow the user to pass a `None` for evalFuncs in `writeSensFile` which will trigger dummy values to be written?

sean-engelstad: Ok I removed the `writeDummySensFile` routine and changed it so that evalFuncs=None triggers the dummy sens file.

sean-engelstad:
* Just changed it so that only the current shape variable it is changing in the animation method is active, this way the mesh chain rule product is only computed for the current shape variable (much faster for multi-shape variable designs).

timryanb (approved):

PR #268, opened 2023-10-30
Fix issues with nonlinear problems when using MPhys wrapper

A-CGray:
- Nonlinear solves are failing when run through the TACS MPhys wrapper
- For example, nonlinear cantilever works fine when run directly with pyTACS, but if I setup the same problem through MPhys, the first linear solve produces NaN's:
```
+-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
|  Iter   |    Time     |  Increment  |     Lambda      |  SubIter  |  Lin iters  |     Lin res     |    Res norm     |  Rel res norm   |     U norm      |     LS step     |  LS iters  |    Flags     |
+-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
|      0  |  0.000e+00  |        0    |   2.000000e-01  |       0   |      -      |        -        |   5.656854e-01  |   1.000000e+00  |   0.000000e+00  |        -        |     -      |              |
|      1  |  2.081e-03  |        0    |   2.000000e-01  |       1   |        1    |           nan   |           nan   |           nan   |           nan   |           nan   |      25    |      PD      |
```
- I think the might be being caused by the fact that we gave the nonlinear solvers their own linear solvers. But when the static problem create's a new matrix and preconditioner we don't pass those on to the nonlinear solver linear solvers.

This PR addresses this by passing the new stiffness matrix and preconditioner to the nonlinear solvers' linear solvers.

I also added a version of the nonlinear hemispherical shell integration test that uses the MPhys wrapper so that using the nonlinear solver through MPhys is covered in the tests. As part of this I found that getting accurate complex-step derivatives was very sensitive to how the solver converged. I therefore added an option to the static problem class that tells TACS to reset the state variables before every solve, which seemed to fix the issue.

Finally I altered the mphys buckling integration test to stop it writing out loads of .f5 files


A-CGray:
@timryanb , took me a while but I finally got this sorted!

tacs/problems/static.py
```diff
@@ -31,6 +31,11 @@ class StaticProblem(TACSProblem):
     defaultOptions = {
         "outputDir": [str, "./", "Output directory for F5 file writer."],
         # Solution Options
+        "resetBeforeSolve": [
+            bool,
+            False,
+            "Reset the states before every solve, this can be useful for avoiding issues when testing derivatives with dinite-difference/complex-step",
```
timryanb, line 37: Small typo in "finite"

timryanb (approved):
Other than that, this looks good to me

A-CGray:
Thanks @timryanb , are you happy to merge this right away or do you want to wait for a release like #319 ?

timryanb:
> Thanks @timryanb , are you happy to merge this right away or do you want to wait for a release like #319 ?

I'm fine merging now if there aren't any breaking changes

PR #269, opened 2023-10-30
Corrections to Caps2tacs docs and support for funtofem shape variable animations

sean-engelstad:
* Correct caps2tacs documentations typos and formatting issues
* Optional dependency with funtofem to animate shape changes for full vehicle models in ESP/CAPS, TACS, and FUNtoFEM using FUNtoFEM shape variables

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -104,6 +104,7 @@ optional_dependencies = {
     "testing": ["testflo>=1.4.7"],
     "docs": ["sphinx", "breathe", "sphinxcontrib-programoutput"],
     "mphys": ["mphys>=1.1.0", "openmdao>=3.25.0"],
+    "f2f-caps-animation" : ["funtofem>=0.3.2", "imageio>=2.16.1"],
 }
 
 # Add an optional dependency that concatenates all others
--- a/setup.py
+++ b/setup.py
@@ -104,7 +104,7 @@ optional_dependencies = {
     "testing": ["testflo>=1.4.7"],
     "docs": ["sphinx", "breathe", "sphinxcontrib-programoutput"],
     "mphys": ["mphys>=1.1.0", "openmdao>=3.25.0"],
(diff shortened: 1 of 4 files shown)


tacs/caps2tacs/gif_writer.py
```diff
@@ -7,15 +7,16 @@ class GifWriter:
...
+        # duration of each frame
+        self._duration = duration
 
     def __call__(self, gif_filename: str, path: str):
         """
         call on current path to create gif of given filename
         """
         gif_filepath = os.path.join(path, gif_filename)
-        with imageio.get_writer(gif_filepath, mode="I", fps=self._fps) as writer:
+        with imageio.get_writer(gif_filepath, mode="I", duration=self._duration) as writer:
```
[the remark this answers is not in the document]
sean-engelstad: Ok I went ahead and added a `caps2tacs` optional dependency section. I put funtofem in there even though caps2tacs can run without it.

setup.py
```diff
@@ -104,6 +104,7 @@ def get_mpi_flags():
     "testing": ["testflo>=1.4.7"],
     "docs": ["sphinx", "breathe", "sphinxcontrib-programoutput"],
     "mphys": ["mphys>=1.1.0", "openmdao>=3.25.0"],
+    "caps2tacs": ["funtofem>=0.3.2", "imageio>=2.16.1"],
```
sean-engelstad, line 107: Not sure if funtofem should really be included as a depency when using caps2tacs. It can run without it, but it does have some additional features if it is available (uses importlib to run without it).

timryanb: @sean-engelstad Let's leave it off the list. As long as there is a guard to prevent an `ImportError`, I'm alright with it

sean-engelstad: @timryanb I removed the funtofem dependency here on the TACS side. Funtofem is modular in that it uses TACS but only through importlib guards. So I think we're good. The funtofem setup.py does not mention tacs.

timryanb (approved):

PR #271, opened 2023-11-02
Bug fix for c++ compiler version and <typeinfo>

bburke38:
- Add standard version for c++ in `Makefile.in.info` and `setup.py`.
- Add `<typeinfo>` header to `TACSShellElement.h`

### DIFF
--- a/Makefile.in.info
+++ b/Makefile.in.info
@@ -18,8 +18,8 @@ MAKE = make -j 8
 AR_FLAGS = rcs
 
 # Flags for debugging and regular compilation versions
-EXTRA_DEBUG_CC_FLAGS = -fPIC -g -std=c++11
-EXTRA_CC_FLAGS = -fPIC -O3 -std=c++11
+EXTRA_DEBUG_CC_FLAGS = -fPIC -g
+EXTRA_CC_FLAGS = -fPIC -O3
 
 # Use this if you have problems with mpich
 # TACS_DEF = -DMPICH_IGNORE_CXX_SEEK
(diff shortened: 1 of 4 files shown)


Makefile.in.info
```diff
@@ -18,8 +18,8 @@ MAKE = make -j 8
 AR_FLAGS = rcs
 
 # Flags for debugging and regular compilation versions
-EXTRA_DEBUG_CC_FLAGS = -fPIC -g
-EXTRA_CC_FLAGS = -fPIC -O3
+EXTRA_DEBUG_CC_FLAGS = -fPIC -g -std=c++11
+EXTRA_CC_FLAGS = -fPIC -O3 -std=c++11
```
timryanb, line 23: We should move this out of the Makefile.in and instead append it in TACS_Common.mk, since this isn't really optional. We're already hard-coding this in the setup.py, so it makes sense to hard-code it on the c++ makefile side as well

bburke38:
Build check failed, error message says c++14 is required.

timryanb (approved):

timryanb:
@bburke38 does this fix the issue on both your compilers?

timryanb:
@bburke38 , just had a conversation with Graeme. We're going to investigate what changes we'd have to make to the code to be able to compile under older c++ standards (c++99 or c++11). I'm going to make this PR a draft until then

bburke38:
> @bburke38 does this fix the issue on both your compilers?

Yes, the commits I made fix the issue on both compilers.

timryanb:
Moving to PR #275

PR #276, opened 2023-11-04
caps2tacs - Parallel Instances of tacsAIMs for huge speed up of post_analysis() for shape optimization

sean-engelstad:
*  active_procs list in the `TacsModel.build` arg is used to create parallel instances of tacs_aim and egads_aim objects from pyCAPS or ESP/CAPS packages.
* When shape variables change, the geometry of each parallel tacsAIM is updated - especially the root tacsAIM (active_procs[0]) which is responsible for the bdf/dat file aka structure mesh.
* The primary reason for creating parallel instances of tacsAIM objects is that the `tacs_aim.post_analysis()` call scales with the # of shape variables and is really slowing down the shape optimization of an aerodynamic vehicle geometry in FUNtoFEM. Namely, the tacsAIM.post_analysis takes 26 minutes for the application case in FUNtoFEM with 8 shape variables and a 40 min CFD analysis. We can achieve a theoretical 8x speedup on this example by computing each shape derivative calculation on a parallel instance of a tacsAIM.
* A new example `7_multiproc_post_analysis.py` was written to demonstrate the speedup of using parallel tacsAIMs on a simple wing geometry. There were 4 shape variables and the theoretical 4x speedup by using 4 parallel tacsAIMs was achieved!
* Much of the underlying structure of the `tacsAIM` class was redone to accomodate parallel instances of the tacsAIM objects. Therefore, all of the ESP/CAPS unittests were ran again and verified. 
* The structure of the `writeSensFile` had to be rewritten since it was never tested before on MPI calls and it was discovered that the `tacs_sens` - Xpts derivatives are distributed across each of the processors and need to be aggregated among them. 
* The `test_caps_shape_derivatives.py` unittest would previously not pass if multiple shape variables were used even though they worked individually (which doesn't make sense). A bug was corrected in which a previous temp variable was accidentally reused. The unittest now works for multiple shape variables and analysis functions.

### DIFF
--- a/tacs/pymeshloader.py
+++ b/tacs/pymeshloader.py
@@ -1210,17 +1210,19 @@ class pyMeshLoader(BaseUI):
         """
 
         self._nastranToLocalNodeIDMap()
-        local_maps = self.comm.gather(self._local_map, root=0)
-        full_map_list = []
-        for local_map in local_maps:
-            full_map_list += local_map
         all_struct_ids = None
+        local_maps = self.comm.gather(self._local_map, root=0)
         if self.comm.rank == 0:
-            all_struct_ids = np.zeros((self.bdfInfo.nnodes), dtype=int)
-            for map in full_map_list:
-                for key in map:
-                    all_struct_ids[int(key)] = map[int(key)]
-            all_struct_ids = list(all_struct_ids)
+            full_map_list = []
+            for local_map in local_maps:
+                full_map_list += local_map
+            all_struct_ids = None
+            if self.comm.rank == 0:
+                all_struct_ids = np.zeros((self.bdfInfo.nnodes), dtype=int)
+                for map in full_map_list:
+                    for key in map:
+                        all_struct_ids[int(key)] = map[int(key)]
+                all_struct_ids = list(all_struct_ids)
         # broadcast to other procs
         all_struct_ids = self.comm.bcast(all_struct_ids, root=0)
         return all_struct_ids
--- a/tacs/pymeshloader.py
+++ b/tacs/pymeshloader.py
@@ -1217,16 +1217,11 @@ class pyMeshLoader(BaseUI):
         all_struct_ids = None
         local_maps = self.comm.gather(self._local_map, root=0)
         if self.comm.rank == 0:
-            full_map_list = []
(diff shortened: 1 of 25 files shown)


tacs/pytacs.py
```diff
@@ -856,7 +856,10 @@ def _checkNonlinearity(self) -> bool:
         # Check if (res2-res0) - 2 * (res1 - res0) is zero (or very close to it)
         resNorm = np.real(res1.norm())
         res2.axpy(-2.0, res1)
-        return (np.real(res2.norm()) / resNorm) > self.getOption("linearityTol")
+        if resNorm == 0.0: # degenerate case
+            return False
+        else:
+            return (np.real(res2.norm()) / resNorm) > self.getOption("linearityTol")
```
timryanb, line 862: Can we just change these lines to improve readability:

```
        if resNorm == 0.0 or (np.real(res2.norm()) / resNorm) <= self.getOption("linearityTol")
            return False
        else:
            return True
```

tacs/pymeshloader.py
```diff
@@ -1210,17 +1214,19 @@ def allLocalNodeIDs(self):
...
-            all_struct_ids = np.zeros((self.bdfInfo.nnodes), dtype=int)
-            for map in full_map_list:
-                for key in map:
-                    all_struct_ids[int(key)] = map[int(key)]
-            all_struct_ids = list(all_struct_ids)
+            full_map_list = []
+            for local_map in local_maps:
+                full_map_list += local_map
+            all_struct_ids = None
+            if self.comm.rank == 0:
```
timryanb, line 1224: This line seems redundant, you already do this check in line 1210

timryanb, line 1224: This line seems redundant, you already do this check in line 1210

tacs/pytacs.py
sean-engelstad: I applied this change now.

tacs/pymeshloader.py
sean-engelstad: I fixed this now

sean-engelstad:
Let me test this out in funtofem again before merging it, if you approve these changes.

sean-engelstad:
@timryanb it passed the ESP/CAPS tests in funtofem now.

timryanb (approved):

PR #278, opened 2023-11-06
Initialise random number generator in `_checkNonlinearity`

### LINKED ISSUE #277: Seeding random number generator in checkNonLinear routine
The checkNonLinearRoutine uses a RNG to initialize a state variable vector to use in its procedure. This RNG isn't seeded and can therefore result unpredicted and unrepeatable behavior (even rarely), this can make debugging intermittently failing unit tests very difficult. To fix this the RNG should be seeded before being used in the routine, see the [element tests](https://github.com/smdogroup/tacs/blob/master/tests/constitutive_tests/test_basic_beam_constitutive.py#L60) for example.

A-CGray:
Closes #277

### DIFF
--- a/src/bpmat/KSM.h
+++ b/src/bpmat/KSM.h
@@ -108,6 +108,7 @@ class TACSVec : public TACSObject {
   // Additional useful member functions
   // ----------------------------------
   virtual void setRand(double lower = -1.0, double upper = 1.0) {}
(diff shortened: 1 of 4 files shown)


A-CGray:
> Probably quickest issue turnaround ever @A-CGray

Spoke too soon

PR #282, opened 2023-11-29
Mphys tweaks

A-CGray:
Adds two things to the TACS MPhys wrapper:

1. A `res_ref` input argument to the TACSBuilder to specify a reference residual value that will be [used by OpenMDAO to scale the residual](https://openmdao.org/newdocs/versions/latest/features/core_features/working_with_components/scaling.html?highlight=res_ref#specifying-a-scaler-on-a-residual)
2. Adds `writeSolution` method to the mphys wrapper to allow users to write out solutions. This is useful when you want to be able to write out solution files but not necessarily after every solve.

Warning: as explained in https://github.com/OpenMDAO/OpenMDAO/issues/3072 OpenMDAO doesn't currently apply the residual scaling correctly to matrix-free partial derivatives, so any check_partials currently fails if you do any residual scaling. The total derivatives are fine though.

### DIFF
--- a/tacs/mphys/postcoupling.py
+++ b/tacs/mphys/postcoupling.py
@@ -130,3 +130,6 @@ class TacsPostcouplingGroup(om.Group):
                 con_group.add_subsystem(
                     constraint.name, con_comp, promotes_inputs=promotes_inputs
                 )
+
+    def writeSolution(self):
+        self.eval_funcs.writeSolution()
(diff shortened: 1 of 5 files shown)


tacs/mphys/postcoupling.py
```diff
@@ -130,3 +130,6 @@ def setup(self):
                 con_group.add_subsystem(
                     constraint.name, con_comp, promotes_inputs=promotes_inputs
                 )
+
+    def writeSolution(self):
```
timryanb, line 134: Can we change this to `write_solution` to match with the mphys/openmdao convention

A-CGray: Done

timryanb (approved):

PR #286, opened 2023-12-11
Modifying BladeStiffenedShellConstitutive class to require fewer inputs

timryanb:
- Adding default inputs to BladeStiffenedShellConstitutive to make initializing easier


A-CGray (approved):
I'd say we require a version bump after this as it's technically a backwards incompatible change right?

PR #289, opened 2024-01-05
Removing support for Python 3.8

timryanb:
- Python 3.8 will reach end of life in October. While there's still a couple months left until this happens some of our dependencies (numpy and pynastran) have already dropped support for it
- I think we should consider removing support for 3.8 early, as maintaining it is becoming more difficult due to the point above.
- This PR will remove support for Python 3.8 and makes the following changes:
  - `oldest-supported-numpy` is being deprecated since numpy 1.25+ now defaults to using a consistent C-API that should always be backwards compatible (see reales notes [here](https://numpy.org/doc/stable/release/1.25.0-notes.html#compiling-against-the-numpy-c-api-is-now-backwards-compatible-by-default))
  - pyNastran released a new update this week which fixes several long-standing errors with processing dynamic loads. I've pinned this version as the new minimum in the setup.py
  - CI Python version has been bumped to 3.9


A-CGray (approved):
This looks good to me, is it waiting onanything @timryanb ?

Also, can we make a new release when this is merged?

timryanb:
> This looks good to me, is it waiting onanything @timryanb ?
> 
> Also, can we make a new release when this is merged?

I'd like to get @gjkennedy approval before merging this in. I'd also like to give people a heads up at the group meeting as I don't know if any of our developers are on 3.8

gjkennedy (approved):

timryanb:
@A-CGray, yes I agree that we should do version update after merging this in. We should also merge any other outstanding PRs with this as well. We can discuss more at the dev meeting

PR #291, opened 2024-01-26
Enable bdf writing for `BladeStiffenedShellConstitutive`

A-CGray:


tacs/pytacs.py
```diff
@@ -1822,47 +1822,54 @@ def writeBDF(self, fileName, problems):
...
+                        matObj = [matObj]
+                    for mat_i in matObj:
+                        if mat_i not in matObjs:
+                            matID = len(matObjs) + 1
+                            mat_i.setNastranID(matID)
+                            matObjs.append(mat_i)
+                        else:
+                            matID = mat_i.getNastranID()
+                        propToMatIDMap[propID].append(matID)
+
```
timryanb, line 1858: Could you add a couple of comments explaining what's going on here?

timryanb:
It looks like these changes are causing the write_bdf tests to hang. Any idea what's going on here?

tacs/constitutive.pyx
```diff
@@ -478,23 +478,42 @@ cdef class MaterialProperties:
...
+        self.E2 = E2
+        self.G23 = G23
+        self.E3 = E3
+        self.rho = rho
+        self.nastranID = 0
+
+    def generateBDFCard(self):
+        return nastran_cards.materials.MAT2(self.nastranID, np.real(self.E1), np.real(self.G12), np.real(self.G13),
+                                            np.real(self.E2), np.real(self.G23), np.real(self.E3), np.real(self.rho))
+
```
timryanb, line 499: Do we need to make a new class for this? It seems we could just re-use the `MaterialProperties` class?

timryanb: Nevermind, I see this is because the class needs MAT2 cards specifically, since the stiffnesses are anisotropic

A-CGray:
@timryanb are you happy with the basic idea behind the current implementation? If so I can go ahead and alter the other constitutive class `generateBDFCard` methods to accept the `mat_ids` input.

PR #297, opened 2024-03-06
Fix issue with delayed transient loads

A-CGray:
Prior to version 1.4.0, pyNastran's `get_load_at_time` was not correctly interpolating transient loads defined using `TLOAD1` when a delay was used. 

For example, `test_trans_beam.py` uses a tip load that is supposed to ramp from 0 to 5N over 1 second with a delay of 1 second. Due to the error in pyNastran, the load was instead being applied as a constant 5N from 0 to 1 seconds and 0N from 1 to 2 seconds, resulting in this solution:

Wrong

With pyNastran 1.4.0, the correct solution is:
Right

This PR bumps the minimum required pyNastran version to 1.4.0 and updates the reference values for the fixed transient integration test.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -129,7 +129,7 @@ setup(
         "numpy<2.0.0",
         "mpi4py>=3.1.1",
         "scipy>=1.2.1",
-        "pynastran>=1.3.3",
+        "pynastran>=1.4.0",
         "numba",
     ],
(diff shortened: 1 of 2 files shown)


A-CGray:
Sorry @timryanb , just saw this is already addressed by #289 , closing.

Issue #301, opened 2024-03-14
Eigenvalue solver tests failing intermittently


timryanb:
One of the integration tests that relates to the eigenvalue solver ( `test_shell_plate_modal.py`) seem to fail intermittently. See these runs ([here](https://github.com/smdogroup/tacs/actions/runs/8273593715/job/22637626137#step:8:18) and [here](https://github.com/smdogroup/tacs/actions/runs/8117679727/job/22190410834#step:8:18)) for example. 

The current theory is that the sporadic nature of the failure is being caused by the random number generator used in the GSEP algorithm [here](https://github.com/smdogroup/tacs/blob/master/src/bpmat/GSEP.cpp#L487).  Seeding the random number generator before running these tests may be a good way of fixing the issue.

PR #302, opened 2024-03-14
Adding random generator seed to all pytacs/mphys tests

timryanb:
Addresses #301. Should fix intermitent test failures we've been seeing on CI recently

### DIFF
--- a/tests/integration_tests/pytacs_analysis_base_test.py
+++ b/tests/integration_tests/pytacs_analysis_base_test.py
@@ -5,6 +5,7 @@ import unittest
 import numpy as np
 from mpi4py import MPI
 
+from tacs import elements
 from tacs import TACS
 from tacs import problems
 from tacs import constraints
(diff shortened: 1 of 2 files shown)


timryanb:
Ran the CI 10 times in a row without any tests failing. The issue seems to be resolved for now

tests/integration_tests/pytacs_analysis_base_test.py
```diff
@@ -63,6 +64,9 @@ def setUp(self):
             # Populate fd/cs perturbation vectors based on user-defined method
             self.setup_tacs_vecs(self.fea_assembler, self.dv_pert, self.xpts_pert)
 
+            # Seed random number generator in tacs for consistent test results
+            elements.SeedRandomGenerator(0)
```
A-CGray, line 68: How does seeding the random generator in `elements` affect the random generator used by the Eigensolver? Are they the same?

timryanb: It calls the same C++ seed function [srand](https://cplusplus.com/reference/cstdlib/srand/) that should seed all random number functions used in tacs.

PR #304, opened 2024-03-22
Displacement Control Fix and Complex Step option for G matrix

sean-engelstad:


sean-engelstad:
I think the change to fix nonzero displacement BCs might have messed up the adjoint solve..

src/TACSBuckling.cpp
```diff
@@ -239,6 +239,10 @@ void TACSLinearBuckling::solve(TACSVec *rhs, TACSVec *u0, KSMPrint *ksm_print) {
       assembler->applyBCs(path);
     } else {
       pc->factor();
+
+      // set BCs associated with displacements (displacement control)
+      assembler->setBCs(path);
+      assembler->setVariables(path);
```
sean-engelstad, line 245: Lines 244-245 are the offending lines that when commented out => the adjoint is correct again and the derivative tests pass.

timryanb: Are these lines actually needed? It seems like the solve should run fine w/o it

timryanb (changes requested):

src/TACSBuckling.cpp
sean-engelstad: The nonzero displacement constraints do not work for the buckling without these lines. Especially for shear buckling

timryanb: They should. This is only changing the initial guess for the linear that happens in 256. Which means these values get overwritten anyways

timryanb: The update is probably not being applied correctly after the solve happens

sean-engelstad: Ok I'll check it out again

timryanb: @sean-engelstad have a look at this fix: https://github.com/sean-engelstad/tacs/pull/1

examples/buckling_plate/2_verification_cases.py
timryanb, line 1: Let's go ahead and make this into an integration test, we have none right now that test dirichelet bcs w/ the buckling solver. Coursen down the mesh and use [test_shell_plate_buckling.py](https://github.com/smdogroup/tacs/blob/master/tests/integration_tests/test_shell_plate_buckling.py) as a starting point/reference.

src/elements/TACSElement.h
```diff
@@ -224,6 +230,20 @@ class TACSElement : public TACSObject {
...
+    @param complexStepFlag The flag for turning on complex step
+  */
+  void setComplexStepGmatrix(bool complexStepFlag) {
+    complexStepGmatrix = complexStepFlag;
+  };
+
+  /**
+    get complexStep mode for Gmatrix (default false)
+  */
+  bool getComplexStepGmatrix() {return complexStepGmatrix;};
```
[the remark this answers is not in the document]
sean-engelstad: @timryanb I was getting this error when trying to call the `elements[i]->setComplexStepGmatrix(True)` in the TACSAssembler. Is there a way I can static_cast the elements to the shell elements?
```
/home/sengelstad6/git/tacs/src/TACSAssembler.cpp:6074:18: error: 'class TACSElement' has no member named 'setComplexStepGmatrix'; did you mean 'bool TACSElement::complexStepGmatrix'? (not accessible from this context)
 6074 |     elements[i]->setComplexStepGmatrix(flag);
      |                  ^~~~~~~~~~~~~~~~~~~~~
In file included from /home/sengelstad6/git/tacs/src/TACSAuxElements.h:26,
                 from /home/sengelstad6/git/tacs/src/TACSAssembler.h:30,
                 from /home/sengelstad6/git/tacs/src/TACSAssembler.cpp:19:
/home/sengelstad6/git/tacs/src/elements/TACSElement.h:754:8: note: declared private here
  754 |   bool complexStepGmatrix = false;
```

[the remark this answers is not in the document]
sean-engelstad: I finished fixing this @timryanb

tests/integration_tests/test_shell_plate_dirichlet_buckling.py
timryanb, line 1: This test is doing too much (mesh writing). Can you you break this into several smaller tests (one for each test case ) using the pre-existing pyTACS test format? Create a static bdf mesh for each test, use [test_shell_plate_buckling.py](https://github.com/smdogroup/tacs/blob/master/tests/integration_tests/test_shell_plate_buckling.py) as starting point for each mesh file. You just need to replace the pytacs problem setup for each case along with the reference quantities to test against. The pyTACS bas test class will handle setting up the actual tests. There also needs to be a description at the top of each test explaining what it does/test

timryanb (changes requested):

tests/integration_tests/test_shell_plate_dirichlet_buckling.py
sean-engelstad: @timryanb I added two unittests for the pure shear and then one with axial + shear combined loading for unstiffened plate buckling. I renamed the previous plate buckling example as *_axial.py to make it more clear how it is different from the other buckling tests. Also, I added a private attribute self._absolute_compare (default False) to the pytacs analysis base test since the shear eigenvalues and derivatives often mode switch (as they only differ at the 7th or 8th digit). This way they only compare the absolute value of the FUNC_REFS to the observed functions (and same for the derivatives since mode switching affects the eigenvalue derivatives too). Not sure if that is the cleanest way to do it, lmk what you think. Also, I shrunk the mesh size to 5x5 for the two new buckling unittests so they don't take up much runtime.

sean-engelstad: Also clang-format is still not working for me at the moment.

src/elements/shell/TACSShellElement.h
```diff
@@ -90,6 +90,12 @@ class TACSShellElement : public TACSElement {
     return quadrature::getQuadratureWeight(n);
   }
 
+  void setComplexStepGmatrix(bool complexStepFlag) {
+    complexStepGmatrix = complexStepFlag;
```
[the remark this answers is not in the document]
timryanb: @sean-engelstad, can you address this?

src/elements/TACSElement.h
```diff
@@ -56,6 +56,16 @@ class TACSElement : public TACSObject {
   */
   int getComponentNum() { return componentNum; }
 
+  /**
+   * get average stresses in the element
+   */
```
timryanb, line 61: Follow docstring format

src/elements/shell/TACSShellElement.h
```diff
@@ -1251,6 +1288,101 @@ void TACSShellElement<quadrature, basis, director, model>::
...
+    model::template computeTyingStrain<vars_per_node, basis>(Xpts, fn, vars, d,
+                                                             ety);
+
+    TacsScalar loc_avgStresses[9];
+    // memset(loc_avgStresses,0,9);
+    for (int i = 0; i < 9; i++) {
+      loc_avgStresses[i] = 0.0;
+    }
+
+    // printf("num vis nodes = %d\n", num_vis_nodes);
```
timryanb, line 1330: @sean-engelstad , can you clean up all o these commented out printf statements?

timryanb (approved):

PR #307, opened 2024-03-26
Fix Nan in Blade Stiffened Shell Constitutive Model

sean-engelstad:
* Getting nans in the `bucklingEnvelopeFailureSens` while running the Mach wing in Funtofem
* N1 << N1crit leads to root becomes 0 due to machine precision. Add small tolerance to prevent divide by zero.


examples/caps_wing/autosave.csm
A-CGray, line 1: Is this file supposed to be included in the PR?

src/constitutive/TACSBladeStiffenedShellConstitutive.cpp
```diff
@@ -2780,17 +2780,17 @@ TacsScalar TACSBladeStiffenedShellConstitutive::bucklingEnvelopeSens(
   TacsScalar root = sqrt(N1Term * N1Term + 4.0 * N12Term * N12Term);
```
A-CGray, line 2780: Since you're adding 1e-13 to the root in all the equations below, you could just add it here instead. Like `TacsScalar root = max(sqrt(N1Term * N1Term + 4.0 * N12Term * N12Term), 1e-13);`

A-CGray (changes requested):

examples/caps_wing/autosave.csm
sean-engelstad: Nope, I'll remove it

sean-engelstad:
@A-CGray - I made the changes you requested

A-CGray (approved):
All good (assuming the tests pass)

src/constitutive/TACSBladeStiffenedShellConstitutive.cpp
```diff
@@ -2778,6 +2778,9 @@ TacsScalar TACSBladeStiffenedShellConstitutive::bucklingEnvelopeSens(
   TacsScalar N1Term = N1 / N1Crit;
   TacsScalar N12Term = N12 / N12Crit;
   TacsScalar root = sqrt(N1Term * N1Term + 4.0 * N12Term * N12Term);
+  if (root == 0.0) {
```
timryanb, line 2781: There should be a `TacsRealPart` around root here

timryanb (approved):

PR #315, opened 2024-05-17
Updating Mphys wrapper for 2.0 update

timryanb:
- Refactored builder class to use new mphys import procedures and variable naming convention
- Changing how coupled tacs problems are setup in builder. No longer using a `coupled` boolean flag, instead user passes forces to be added to RHS as list of str
- MPhys examples and tests in tacs repo are updated and working (using this [PR](https://github.com/OpenMDAO/mphys/pull/176) of MPhys)
- CI expected to fail until MPhys 2.0 is released 
- To be merged in when after MPhys release


setup.py
```diff
@@ -104,7 +104,7 @@ def get_mpi_flags():
 optional_dependencies = {
     "testing": ["testflo>=1.4.7"],
     "docs": ["sphinx", "breathe", "sphinxcontrib-programoutput"],
-    "mphys": ["mphys>=1.1.0,<2.0.0", "openmdao>=3.25.0"],
+    "mphys": ["mphys>=2.0.0b1", "openmdao>=3.25.0"],
```
A-CGray, line 107: We should remove the `b1` now right?

A-CGray (approved):
I've run the integration tests locally with mphys 2.0 and all are fine, so once mphys-2.0.0 is released on pypi and the CI jobs pass I'm happy to merge

PR #319, opened 2024-06-10
Blade stiffened shell improvements

A-CGray:
This PR contains various improvements to the `TACSBladeStiffenedShellConstitutive` constitutive class:

1. Adds stiffener column buckling and crippling failure modes
2. Material failure criterion is now computed at bottom as well as the top of the stiffener
3. Adds capability to choose which of the failure modes are used in the overall failure calculation (material failure, global panel buckling, inter-stiffener/local panel buckling, stiffener column buckling, stiffener crippling). At the python level, these can be set through the `setFailureModes` method
4. Related to 3, the unit tests for this constitutive model now contain separate tests for each failure mode's derivatives as well as tests for the full combined failure envelope
5. Rather than having them hardcoded in the `evalFailure` and failure sensitivity methods, I have moved the global and local panel buckling calculations to their own methods. This should reduce the amount of code duplication required in @sean-engelstad 's #311 PR


src/constitutive/TACSBladeStiffenedShellConstitutive.h
```diff
@@ -142,13 +154,82 @@ class TACSBladeStiffenedShellConstitutive : public TACSShellConstitutive {
       int _stiffenerHeightNum, TacsScalar _stiffenerThick,
       int _stiffenerThickNum, int _numStiffenerPlies,
       TacsScalar _stiffenerPlyAngles[], TacsScalar _stiffenerPlyFracs[],
-      int _stiffenerPlyFracNums[], TacsScalar _flangeFraction = 1.0);
+      int _stiffenerPlyFracNums[], TacsScalar _flangeFraction = 1.0,
+      bool _includePanelMaterialFailure = true,
+      bool _includeStiffenerMaterialFailure = true,
+      bool _includeGlobalBuckling = true, bool _includeLocalBuckling = true,
+      bool _includeStiffenerColumnBuckling = true,
+      bool _includeStiffenerCrippling = true);
```
timryanb, line 162: The function handle for this class is starting to get really long. Is there a way we can condense some of this?
Maybe be converting each of these individual flags to a single binary composite flag, similar to how we handle TACS_OUTPUT (see [here](https://github.com/smdogroup/tacs/blob/master/examples/crm/crm.py#L199-L207)).

tests/integration_tests/test_shell_blade_stiffened_plate_quad.py
```diff
@@ -42,7 +42,7 @@ class ProblemTest(PyTACSTestCase.PyTACSTest):
         "point_load_cgy": 0.500000000000004,
         "point_load_cgz": -0.0035714285714285718,
         "point_load_compliance": 71.87161795633577,
-        "point_load_ks_vmfailure": 0.3304517634314118,
```
timryanb, line 45: What's the reason for this large jump in failure values?

src/constitutive/TACSBladeStiffenedShellConstitutive.h
A-CGray: I could just remove these arguments from the constructor and have the setter methods as the only way to enable and disable different failure modes, same as is currently done for the DV bounds

timryanb: That'd work too

tests/integration_tests/test_shell_blade_stiffened_plate_quad.py
A-CGray: I guess one of the new failure modes is the active one

timryanb: Can we look a little more into which failure mode is causing the jump? Also have you tried running these changes through a full optimization case? I just want to know that these new checks give reaslistic results

A-CGray: Yeah I've run some optimizations on the MACH tutorial wing with the new failure modes, here's an example
image

As you can see, not much difference in the final design except a few panels with lower stiffener heights (presumably to avoid stiffener crippling).

Let me do a bit of digging to see which mode is critical in that integration test.

A-CGray: Looks like it's the stiffener column buckling that's causing the big rise in this case. If I disable that mode I get failure values slightly higher than the old values and if I also remove the crippling failure mode then I get back the old values pretty much exactly.

timryanb: Ok, that sounds good to me. Since this is a breaking change, we should hold off on merging until we're ready for a new release. Is that alright @A-CGray?

timryanb:
@sean-engelstad, could you review this PR when you get a chance and see if these new methods can be utilized in your GP panel PR #311?

sean-engelstad:
Hi @timryanb I should be able to review it later today

src/constitutive/TACSBladeStiffenedShellConstitutive.cpp
```diff
@@ -2853,6 +3066,552 @@ bool TACSBladeStiffenedShellConstitutive::testBucklingEnvelopeSens(
...
+  // t is the flange thickness
+  //
+  // Here we compute the ultimate compressive load using the same criteria
+  // used to compute the stiffener material failure value but using on the
+  // axial strain component. This gives: SR_Crippling = SR_comp/1.63 *
+  // (b/t)^0.717 Where SR_Crippling is the crippling strength ratio SR_comp is
+  // the strength ratio computed using only the compressive component of the
+  // strain
+
+  const int numPlies = this->numStiffenerPlies;
```
sean-engelstad, line 3222: Have you tested the stiffener crippling critical loads here to ensure that they give reasonable limits on stiffener aspect ratio? Could use a simple flat panel example and axial loads just to check.

tests/integration_tests/test_shell_blade_stiffened_plate_quad.py
sean-engelstad: I recommend implementing a mode to change the output values to f5 so you can see the failure indices. It would actually be better if you move part of my code for this to the base class I think. Then you could see each failure value as a DV temporarily and figure out which one is critical. See line 840 in `src/constitutive/TACSGPBladeStiffenedShellConstitutive` in my PR

sean-engelstad (changes requested):
Sorry didn't mean to click approve on the PR. Want to see printout of the failure values for where stiffener column buckling is active and can you provide the source for the column buckling equations?

src/constitutive/TACSBladeStiffenedShellConstitutive.cpp
```diff
@@ -2646,15 +2705,166 @@ TacsScalar TACSBladeStiffenedShellConstitutive::computeCriticalShearLoad(
...
     const TacsScalar D1, const TacsScalar D2, const TacsScalar D3,
-    const TacsScalar L, TacsScalar* D1Sens, TacsScalar* D2Sens,
-    TacsScalar* D3Sens, TacsScalar* LSens) {
-  TacsScalar L2 = L * L;
-  TacsScalar L3 = L2 * L;
-  TacsScalar D32 = D3 * D3;
-  TacsScalar root = sqrt(D1 * D2);
-  TacsScalar xi = root / D3;
+    const TacsScalar L, TacsScalar* const D1Sens, TacsScalar* const D2Sens,
+    TacsScalar* const D3Sens, TacsScalar* const LSens) {
```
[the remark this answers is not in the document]
sean-engelstad: This is not that important as long as the test works though.

sean-engelstad:
@A-CGray if you add a mode and setter for writing out different values to the f5 file or in the DVs (like what I did in my PR), then you can see the failure values. Also, I think you should change the way I did it and have dv1 be argmax(fails) basically so you can see the index of which failure criterion is maximum.

timryanb:
> @A-CGray if you add a mode and setter for writing out different values to the f5 file or in the DVs (like what I did in my PR), then you can see the failure values. Also, I think you should change the way I did it and have dv1 be argmax(fails) basically so you can see the index of which failure criterion is maximum.



> @A-CGray if you add a mode and setter for writing out different values to the f5 file or in the DVs (like what I did in my PR), then you can see the failure values. Also, I think you should change the way I did it and have dv1 be argmax(fails) basically so you can see the index of which failure criterion is maximum.

@sean-engelstad, there should be a cleaner way of doing this. Maybe allowing for multiple failure values to be returned in the f5 file (similar to how dvs are handled). But writing failure information out to the dv field seems confusing

sean-engelstad:
@timryanb Yeah if there is a way to add extra outputs to the f5 file that say `fail0` or `panelStress`, etc. and then `failIndex` for the argmax(fails) part that would be better.

timryanb:
In that case, let's make this part a separate PR/feature. I'll add it to the issues page

timryanb:
See #320 @sean-engelstad  @A-CGray

src/constitutive/TACSBladeStiffenedShellConstitutive.cpp
```diff
@@ -2853,6 +3066,552 @@ bool TACSBladeStiffenedShellConstitutive::testBucklingEnvelopeSens(
           fabs(dfdN12RelError) < tol && fabs(dfdN12CritRelError) < tol);
 }
 
+TacsScalar TACSBladeStiffenedShellConstitutive::evalStiffenerColumnBuckling(
+    const TacsScalar stiffenerStrain[]) {
+  TacsScalar stiffenerStress[TACSBeamConstitutive::NUM_STRESSES];
+  this->computeStiffenerStress(stiffenerStrain, stiffenerStress);
+  // The first component of the stiffener stress is the axial force
```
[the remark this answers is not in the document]
A-CGray: Have added some more comments explaining the various buckling failure modes now.

This failure mode is for column buckling of the stiffener on it's own, hence no dependence on the skin properties. Column buckling of the combined skin + stiffener section is included in the global panel buckling failure mode

src/constitutive/TACSBladeStiffenedShellConstitutive.cpp
```diff
@@ -1890,21 +1742,32 @@ TacsScalar TACSBladeStiffenedShellConstitutive::computeStiffenerFailure(
...
 
   // Compute the strain state at the tip of the stiffener
-  TacsScalar zTipOffset = -(this->stiffenerHeight + this->stiffenerThick) -
-                          this->computeStiffenerCentroidHeight();
-  TacsScalar tipStrain[3];
-  memset(tipStrain, 0, 3 * sizeof(TacsScalar));
-  tipStrain[0] = stiffenerStrain[0] + zTipOffset * stiffenerStrain[2];
+  TacsScalar zCentroid = this->computeStiffenerCentroidHeight();
+  TacsScalar zTipOffset =
+      -(this->stiffenerHeight + this->stiffenerThick) - zCentroid;
```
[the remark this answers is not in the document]
A-CGray: Because in this model I assume that the stiffener height value is the length of the stiffener web above the flange, not including the flange.  See below:

image

I went back and forth on whether to do it this way or to make the stiffener height the total height from the bottom to top of the stiffener. Either way you do it the equations get a bit tricky somewhere. With this approach the calculation of the stiffener cross-section properties is simpler but you have to include the thickness of the flange when computing offsets like this. If you do it the other way then computing the stiffener cross section properties is more complicated because you have to avoid double counting the material that's in both the flange and the web.

src/elements/shell/TACSShellElement.h
```diff
@@ -1487,20 +1487,14 @@ void TACSShellElement<quadrature, basis, director, model>::getOutputData(
...
-        data[12] = con->evalDesignFieldValue(elemIndex, pt, X, 6);
-        data += 13;
+        for (int failInd = 0; failInd < 7; failInd++) {
+          data[failInd] =
+              con->evalFailureFieldValue(elemIndex, pt, X, e, failInd);
+        }
+        for (int dvInd = 0; dvInd < 7; dvInd++) {
+          data[dvInd + 7] = con->evalDesignFieldValue(elemIndex, pt, X, dvInd);
+        }
+        data += 14;
```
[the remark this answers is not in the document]
A-CGray: Done

src/constitutive/TACSBladeStiffenedShellConstitutive.cpp
```diff
@@ -1890,21 +1742,32 @@ TacsScalar TACSBladeStiffenedShellConstitutive::computeStiffenerFailure(
   TACSOrthotropicPly* ply = this->stiffenerPly;
 
   // Compute the strain state at the tip of the stiffener
-  TacsScalar zTipOffset = -(this->stiffenerHeight + this->stiffenerThick) -
-                          this->computeStiffenerCentroidHeight();
-  TacsScalar tipStrain[3];
-  memset(tipStrain, 0, 3 * sizeof(TacsScalar));
-  tipStrain[0] = stiffenerStrain[0] + zTipOffset * stiffenerStrain[2];
+  TacsScalar zCentroid = this->computeStiffenerCentroidHeight();
```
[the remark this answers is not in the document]
A-CGray: Right, it's because the stiffener is assumed to be oriented opposite from the shell normal, so the offset of the stiffener centroid from the shell mid-plane is negative. I've changed it now so that `computeStiffenerCentroidHeight` and `computeStiffenerCentroidHeightSens` treat the height as a positive value and updated all the places they're used to account for the change.

src/constitutive/TACSBladeStiffenedShellConstitutive.cpp
```diff
@@ -2646,15 +2705,166 @@ TacsScalar TACSBladeStiffenedShellConstitutive::computeCriticalShearLoad(
...
+  }
+}
+
+TacsScalar
+TACSBladeStiffenedShellConstitutive::computeCriticalLocalAxialLoadSens(
+    const TacsScalar D11, const TacsScalar D22, const TacsScalar D12,
+    const TacsScalar D66, const TacsScalar L, TacsScalar* const D11Sens,
+    TacsScalar* const D22Sens, TacsScalar* const D12Sens,
+    TacsScalar* const D66Sens, TacsScalar* const LSens) {
+  const double pi2 = M_PI * M_PI;
```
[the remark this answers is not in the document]
A-CGray: If you look at where this method gets called in `addLocalPanelBucklingDVSens`, the value of `dfN1crit` isn't known yet because it relies on the value of `N1Crit`, which is computed by this method.

I agree that the ordering of operations in the buckling sensitivities could be improved quite a bit, but it doesn't seem to have a significant effect on the overall computation time so I will just leave it as is for now.

src/constitutive/TACSBladeStiffenedShellConstitutive.cpp
```diff
@@ -2172,15 +2081,158 @@ void TACSBladeStiffenedShellConstitutive::
...
+  TacsScalar dfdD1, dfdD2, dfdD3, dfdPanelLength;
+  this->computeCriticalShearLoadSens(D1, D2, D3, L, &dfdD1, &dfdD2, &dfdD3,
+                                     &dfdPanelLength);
+  dfdD1 *= dfdN12CritGlobal;
+  dfdD2 *= dfdN12CritGlobal;
+  dfdD3 *= dfdN12CritGlobal;
+  dfdPanelLength *= dfdN12CritGlobal;
+  // N1CritGlobal = M_PI * M_PI * D1 / (L * L);
+  dfdD1 += dfdN1CritGlobal * (M_PI * M_PI / (L * L));
+  dfdPanelLength += dfdN1CritGlobal * (-2.0 * M_PI * M_PI * D1 / (L * L * L));
```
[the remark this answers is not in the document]
A-CGray: Good idea

src/constitutive/TACSBladeStiffenedShellConstitutive.cpp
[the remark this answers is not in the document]
A-CGray: I think these equations are going to be pretty gnarly however they're written, even the power series method is not so straightforward because all of the equations depend on both D1/2/3 and xi which is a function of D1/2/3 so it's not obvious what the exponent is.

A-CGray:
Just to confirm @timryanb @sean-engelstad , the plan is to merge this PR, then make the necessary updates to #311 and merge that?

sean-engelstad (approved):
This looks very good to me! We may still want to wait on Tim's review and see if he wants us to do some comparisons of predicted buckling loads and optimized designs btw the two classes here.

A-CGray:
Sure, for reference @sean-engelstad , here are the results with my constitutive model:

### Fixed stiffener pitch no stiffener crippling constraint
```
>>>python optimize_stiffened_plate.py 
.
.
.
Objectives
{'CompAndShear.Mass': array([10.79339379])}

Optimization terminated successfully    (Exit mode 0)
            Current function value: 10.793393789903986
            Iterations: 8
            Function evaluations: 10
            Gradient evaluations: 8
Optimization Complete
-----------------------------------
Optimal sizing:
================================
Stiffener pitch: 125.0 mm
Panel thickness: 2.9308217613181076 mm
Skin ply fractions:
[0.13953 0.36047 0.36047 0.13953]
Stiffener height: 115.00000000000001 mm
Stiffener thickness: 1.6478083873616798 mm
Stiffener ply fractions:
[0.59402 0.16239 0.16239 0.0812 ]
```

### Fixed stiffener pitch with stiffener crippling constraint
```
>>>python optimize_stiffened_plate.py --includeStiffenerBuckling
.
.
.
Objectives
{'CompAndShear.Mass': array([10.89286276])}

Optimization terminated successfully    (Exit mode 0)
            Current function value: 10.89286276157974
            Iterations: 13
            Function evaluations: 17
            Gradient evaluations: 13
Optimization Complete
-----------------------------------
Optimal sizing:
================================
Stiffener pitch: 125.0 mm
Panel thickness: 2.948897670317985 mm
Skin ply fractions:
[0.13953 0.36047 0.36047 0.13953]
Stiffener height: 114.20492349292782 mm
Stiffener thickness: 1.6794607381336215 mm
Stiffener ply fractions:
[0.59402 0.16239 0.16239 0.0812 ]
```

### With stiffener pitch DV, no stiffener crippling
```
>>>python optimize_stiffened_plate.py --useStiffPitchDV
.
.
.
Objectives
{'CompAndShear.Mass': array([9.72264797])}

Optimization terminated successfully    (Exit mode 0)
            Current function value: 9.722647972607012
            Iterations: 9
            Function evaluations: 11
            Gradient evaluations: 9
Optimization Complete
-----------------------------------
Optimal sizing:
================================
Stiffener pitch: 159.99999999997792 mm
Panel thickness: 3.701976915981655 mm
Skin ply fractions:
[0.13953 0.36047 0.36047 0.13953]
Stiffener height: 150.00000000000003 mm
Stiffener thickness: 0.8902831226396368 mm
Stiffener ply fractions:
[0.59402 0.16239 0.16239 0.0812 ]
```

### With stiffener pitch DV and stiffener crippling
```
>>>python optimize_stiffened_plate.py --useStiffPitchDV --includeStiffenerBuckling
.
.
.
Objectives
{'CompAndShear.Mass': array([10.88611761])}

Optimization terminated successfully    (Exit mode 0)
            Current function value: 10.886117606613414
            Iterations: 12
            Function evaluations: 17
            Gradient evaluations: 12
Optimization Complete
-----------------------------------
Optimal sizing:
================================
Stiffener pitch: 124.0118771982418 mm
Panel thickness: 2.9291042508631016 mm
Skin ply fractions:
[0.13953 0.36047 0.36047 0.13953]
Stiffener height: 114.01187897381736 mm
Stiffener thickness: 1.6777439670951486 mm
Stiffener ply fractions:
[0.59402 0.16239 0.16239 0.0812 ]
```

sean-engelstad:
Here's what I get with the closed-form solutions in `TacsGPBladeStiffenedShellConstitutive` subclass when I run a copy of your optimization script with stiffener crippling on and no stiffener pitch DV.  

```
Design Vars
{'dvs.dv_struct': array([0.01196416, 0.05817722, 0.00388759])}

Nonlinear constraints
{'CompAndShear.struct_post.eval_funcs.KSFailure': array([1.]),
 'Pressure.struct_post.eval_funcs.KSFailure': array([0.46583251]),
 'Pressure.struct_post.eval_funcs.MaxDispZ': array([0.002])}

Linear constraints
None

Objectives
{'CompAndShear.struct_post.mass_funcs.Mass': array([28.20693225])}

Optimization terminated successfully    (Exit mode 0)
            Current function value: 28.206932245353478
            Iterations: 31
            Function evaluations: 51
            Gradient evaluations: 31
Optimization Complete
-----------------------------------
Optimal sizing:
================================
Stiffener pitch: 125.0 mm
Panel thickness: 11.964156918285498 mm
Skin ply fractions:
[0.13953 0.36047 0.36047 0.13953]
Stiffener height: 58.17722485065222 mm
Stiffener thickness: 3.887590239767674 mm
Stiffener ply fractions:
[0.59402 0.16239 0.16239 0.0812 ]
```

A-CGray:
Thanks @sean-engelstad , can you also try running without stiffener crippling? I have a suspicion that's what's causing most of the difference between the two models here

A-CGray:
For the plate with the default design variables values:
```
Stiffener pitch: 125.0 mm
Panel thickness: 2.1717 mm
Skin ply fractions:
[0.13953 0.36047 0.36047 0.13953]
Stiffener height: 57.0 mm
Stiffener thickness: 6.4772727272727275 mm
Stiffener ply fractions:
[0.59402 0.16239 0.16239 0.0812 ]
```

I get these values for each failure mode in the compression+shear load case:
- Skin failure: 0.202345
- Stiffener failure: 0.079251
- Global panel buckling: 1.332269
- Inter-stiffener buckling: 0.428721
- Stiffener column buckling: 0.292454
- Stiffener crippling: 0.113508

These are the values not including the safety factor

A-CGray:
Update for @timryanb 

- Me and @sean-engelstad compared our failure mode values for the default DVs in this plate case 
- The biggest difference is actually in the predicted critical loads for global and local panel buckling, my critical axial and shear loads for global buckling are almost an order of magnitude higher than Sean's.
- Looking at the equations we use to compute the critical axial load, mine computes the bending stiffness of the skin+stiffener cross section about it's centroid while it looks like the formula Sean is using may not account for the offset of the skin and stiffener from this centroid, which might explain the much lower critical load.
- Another difference is that Sean's equation is for a plate pinned on all edges, while my model uses the Euler column buckling equation, so there is no stiffening from the sides of the panel.
- We tried to figure out whose model was closer to reality by comparing our critical load predictions to TACS eigenvalue buckling analyses of plates with explicitly modelled stiffeners.
- For all the plates we could successfully run eigenvalue analyses on, our models were actually relatively close (e.g within 10%), however these were all with extremely low-aspect ratio (AR=1) stiffeners
- However, we couldn't successfully run eigenvalue analyses on plates with higher stiffener aspect ratios or plates where the stiffeners were a larger fraction of the overall stiffness. For example, we couldn't successfully run an eigenvalue analysis on the plate above where our models differ by an order of magnitude

sean-engelstad:
@timryanb I think I know how to fix this, there is an inconsistency in stiffener and panel D11 centroid in my code. I will also try and put together a higher quality finite element model of the stiffened panel (validated against the literature this week) that we can compare against

timryanb:
Sounds good. Thanks @sean-engelstad

sean-engelstad:
My FEA models of stiffened panels in my ml_buckling repo are working better for stiffener aspect ratios of around 20 now. There was an inconsistency in not using the overall centroid before. And that threw off my shift-and-invert eigenvalue guesses leading me to not be able to run higher stiffener ARs. My closed-form solution is matching better against the FEA models now for larger stiffeners (with 2% error at high ARs). Need to update my TACS code to compute panel Dij about overall centroid and differentiate that step. Then I can post an updated benchmark design of the stiffened panel Ali made

sean-engelstad:
Updated stiffened panel verification case that I put together in my repo, ml_buckling. Here's a panel design and its inputs
```
Stiffened panel geometry object:
        a = 6.25773160690871
        b = 1.0
        h = 0.01
        num_stiff = 1
        stiffener pitch = 0.5
        h_w = 0.07663302688934862 (height of stiffener)
        t_w = 0.003831651344467431 (thick of stiffener)
        AR = 6.25773160690871
        SR = 100.0
        stiff_AR = 20.0
Plate material:
Composite material object
        E11 = 1.38e11
        nu12 = 0.3
        _E22 = 1.38e11
        _G12 = E11 / 2 / (1+nu12)
        _G23 = None
        _G13 = None
        _ply_angles = [0]
        _ply_fractions = [1.0]
        material_name = None
        symmetric = True
        ref_axis = [1 0 0]
Stiffener material:  same as panel
Advanced parameters in the stiffened panel analysis obj
        rho_0 = 5.9999999999306555
        xi plate = 0.9193240803560768
        xi stiff = 0.9999999999999999
        gamma = 10.000000000000062
        delta = 0.058726208102236284
        zeta plate = 0.00028571428571428574
        zeta stiff = 0.007142857142857143
        stiffAR = 20.0
Mesh + Case Settings
        exx = 9.28592e-05
        exy = 0.00000e+00
        eyy = 0.00000e+00
        num nodes = 2803
In plane loads
        Nxx=128145.6574428344
        Nxy=0.0
```

Then here are my results comparing TACS and the FEA model:
Axial
----------------
Predicted by TACS GP closed-form: N11,cr = 1.141e6 N/m
Predicted by FEA:                             N11,cr = 1.0058e6 N/m 
Rel Error: 3.58%

Screenshot from 2024-09-16 20-01-55

Shear
----------------
Predicted by TACS GP closed-form: N12,cr = 1.009e6 N/m
Predicted by FEA:                             N12,cr = 1.018e6 N/m 
Rel Error: 0.9%

image


Here's also a plot I made to compare CF and FEA in my code for a bunch of models with a single stiffener in the middle. The closed-form solutions are most accurate at unstiffened case or high aspect ratios (although shear is only accurate at high and low ARs). For lower ARs, the mode distortion from the stiffener end constraints results in the global mode distorting to a semi global-local mode and modal assurance criterion doesn't pick it up. So I think these are good results (since they match so well at high ARs and the Timoshenko : Theory of Elasticity book makes a similar comment on a panel CF of this sort).
image

timryanb:
Thanks for doing all of this verification work @sean-engelstad. Your results seem to be in good agreement with theory and FEA. Did any of these changes bring your and Ali's results closer together in agreement or are we still waiting on that result?

sean-engelstad:
@timryanb, I made a new comparison case here. So @A-CGray could you change the plate to this design and post your comparison results? I can also do the optimization with the old stiffened panel => but need to update my derivatives in the const class (so may be a few days out from that).

A-CGray:
@sean-engelstad , sorry to be a pain but is it possible to run a buckling analysis on a panel with the same cross section but that's wider than it is long? Or at least one where the edges of the plate parallel to the stiffener aren't constrained? I'm pretty certain my global buckling prediction will not be at all meaningful for this case because it assumes that the panel is infinitely wide, whereas in this case the buckling mode is clearly dictated by the panel width.

sean-engelstad:
This figure that I've added on stiffened panel verification in my paper shows that once `rho_0` the affine aspect ratio is low enough and `gamma` the EI ratio of stiffener to panel is large enough, global to local mode mixing begins. The greater the number of $N_s$ stiffeners there are, the more you can delay the onset of local modes and preserve a mostly global mode shape. For this reason, I've selected a low aspect ratio panel with $N_s = 8$ stiffeners. I've also gone with a stiffener aspect ratio of 5 although I could go a bit higher. I'm getting that the closed-form solution is conservative against the FEA for low aspect ratios, but it is less conservative the closer it is to a [0 deg]_1 laminate ABD matrix (in which case then it matches very closely).

image

```
Stiffened panel geometry object:
        a = 0.20277833751554175
        b = 1.0
        h = 0.01
        num_stiff = 8
        stiffener pitch = 0.1111111111111111
        w_b = 0.0
        t_b = 0.0
        h_w = 0.01797546789982607
        t_w = 0.003595093579965214
        AR = 0.20277833751554175
        SR = 100.0
        stiff_AR = 5.0
Plate material:
Composite material object
        E11 = 138000000000.0
        nu12 = 0.3
        _E22 = 138000000000.0
        _G12 = 53076923076.92307
        _G23 = None
        _G13 = None
        _ply_angles = [0.0]
        _ply_fractions = [1.0]
        material_name = None
        symmetric = True
        ref_axis = [1 0 0]
Stiffener material:
Composite material object
        E11 = 138000000000.0
        nu12 = 0.3
        _E22 = 138000000000.0
        _G12 = 53076923076.92307
        _G23 = None
        _G13 = None
        _ply_angles = [0.0]
        _ply_fractions = [1.0]
        material_name = None
        symmetric = True
        ref_axis = [1 0 0]

Advanced parameters in the stiffened panel analysis obj
        rho_0 = 0.2
        xi plate = 0.9727850217307755
        xi stiff = 1.0
        gamma = 1.0000000000000016
        delta = 0.058161140319181945
        zeta plate = 0.00028571428571428574
        zeta stiff = 0.11428571428571428
        stiffAR = 5.0
```

# Axial Loading:
I think it makes sense that the FEA eigenvalue is quite a bit higher than the closed-form as the buckling mode is significantly more restrained due to the stiffeners.
```
Mesh + Case Settings
        exx = 8.78028e-05
        exy = 0.00000e+00
        eyy = 0.00000e+00
        num nodes = 3057
In plane loads
        Nxx=121167.88013431858
        Nxy=0.0
MAC identified global mode 0 with shape (m,n)=(1,1) and max_similarity=0.9981
Eigenvalues in ml_buckling code:
        my CF min lambda = 51.985570043461585
        FEA min lambda = 88.47617896320556
```


image
Namely the predicted closed-form vs FEA critical loads by TACS are:

A-CGray:
@timryanb , just ran @sean-engelstad 's case, here's what I get:

`N1Crit = 6.188025e+06, N12Crit = 2.037140e+07`

Comparing all 3 methods:

```
    TACS GP closed-form: N11,cr =  6,057,629 N/m  N12,cr = 18,523,089  N/m
    TACS FEA prediction: N11,cr = 10,719,644 N/m  N12,cr = 24,532,852  N/m
TACSBladeStiffenedShell: N11,cr =  6,188,025 N/m  N12,cr = 20,371,400 N/m
```

PR #321, opened 2024-06-26
Adding capability to write multiple failure values to f5 file

### LINKED ISSUE #320: Add ability to write out multiple failure values in f5
Probably something like how dv outputs are currently handled (`failure0`, `failure1`, etc.).

timryanb:
- shell/beam f5 files will now have 5 unique failure fields ("failure0", "failure1", etc.)
- This is handled by new `evalFailureFieldValue` method in `TACSConstitutive` class
- This allows for multiple failure mode contours to be visualized in post-processing (as was requested in PR's #319 and #311)
- "failure0" defaults to whatever is given by `evalFailure` method (same value we've previously been writing out) and the rest default to 0 (This can be overridden at the sub-class level)
- Closes #320


A-CGray (approved):
Tested this out and it works for the blade stiffened shell model

untitled

A-CGray:
Are you happy for me to merge this @timryanb or do you want to wait for another review?

timryanb:
I'm fine with that @A-CGray

PR #323, opened 2024-07-12
Mphys helper functions

A-CGray:
Adds a helper function to the TACS MPhys wrapper for adding constraints. `add_tacs_constraints` takes in an OpenMDAO model and an MPhys scenario and automatically adds any TACS constraints it finds in the scenario as constraints in the model. Saves the user having to manually call `add_constraint` for all of their TACS constraints. 

I put it in a `utils` file for now, happy to take suggestions on a better place for it.

### DIFF
--- a/tacs/mphys/utils.py
+++ b/tacs/mphys/utils.py
@@ -1,3 +1,4 @@
+from functools import wraps
 import warnings
 import tacs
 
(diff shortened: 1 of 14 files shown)


timryanb:
@A-CGray, the log function call looks helpful, but it seems more like a general OpenMDAO feature than specific to TACS. Could we reach out to OpenMDAO/OpenMDAO and see if they'd be willing to adopt this feature (or such a feature already exist)?

A-CGray:
Sure, I'll remove it from this PR and bring it up with the OpenMDAO team

tacs/mphys/utils.py
```diff
@@ -0,0 +1,66 @@
...
+    function yourself.
+
+    This function relies on you having defined the lower and/or upper bounds
+    of the constraints when you created them in your `constraint_setup`
+    function.
+
+    Parameters
+    ----------
+    model : OpenMDAO model
+        Model to add constraints to
```
timryanb, line 18: Is the `model` really needed as an input here? It looks like it's only used to get the root proc for printing? Can we just use the comm from the scenario?

A-CGray: You need the model because that's what the constraints are added to

A-CGray: Oh, can call `add_constraint` on a scenario too? In that case I agree and will remove the model input

timryanb: Ah, I see that now. But you can just add the constraints directly to the subsystem and skip the model. Something like: `constraint.add_constraint(name, equals=lb, linear=True)`

timryanb: Ah, nvm. Beat me to it

timryanb (approved):

A-CGray:
@timryanb looks there is an OpenMDAO bug that results in an indexing error if you only have linear constraints, I'll open an issue on the OpenMDAO repo

PR #337, opened 2024-11-01
Making mpi4py 3.1.5 new minimum version for build

timryanb:
Python 3.12+ seems to be incompatible with older versions of mpi4py. In order to fix this we'll need to bump up minimum mpi4py version to 3.1.5.

### DIFF
--- a/conda/meta.yaml
+++ b/conda/meta.yaml
@@ -37,7 +37,7 @@ requirements:
     - mpich-mpicxx  # [mpi == "mpich"]
     - {{ compiler('cxx') }}
     - make
-    - mpi4py >=3.1.1,<4.0.0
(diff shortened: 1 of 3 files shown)


A-CGray (approved):
Worked fine for me with: 

- OpenMPI 4.1.6
- python 3.11.9
- mpi4py 3.1.6

SeiyonArulampalam:
Got it to work on my end as well using:
- mpi4py=3.1.6
- python 3.10.12

bburke38:
Seems to have installed correctly on my Mac. It updated mpi4py in the process to mpi4py==3.1.6. This was using python==3.9.16 (I am using an older version of Python for compatibility with something else).

PR #342, opened 2024-12-09
Version bump 3.8.0

timryanb:
This version bump updates TACS MPhys wrapper to support MPhys 2.0. 
This will be a backwards-incompatible update for any MPhys users.
To be merged in after PR #315 and #337


sean-engelstad (approved):
mpi4py-related version bump is verified backwards compatible with older python versions

PR #348, opened 2025-01-16
renaming `check_partials` attribute in mphys components

timryanb:
The following [PR](https://github.com/OpenMDAO/OpenMDAO/pull/3424) in OpenMDAO will add a `check_partials` method to all OpenMDAO components. I've refactored the already existing `check_partials` flag in some mphys components in order to prevent a conflict.


timryanb:
> Kinda pedantic but, to me, `check_partials_mode` sounds like a variable that should define what kind of partials check is being done, not whether the partials are being checked or not. Maybe we could use the name `under_check_partials` to mirror the `under_complex_step` that OpenMDAO already uses?

Should be modified now

A-CGray (approved):

PR #354, opened 2025-01-28
Adding analytic geometric stiffness for TACSShellElement

### LINKED ISSUE #246: Expose finite difference step size as an option in buckling problem

### LINKED ISSUE #294: Implement analytic geometric stiffness calculation for shell elements
The geometric stiffness matrix used in the shell element class is currently approximated by finite differencing along the path of the nonlinear stiffness matrix. This leads to some error, loss in robustness, and speed penalty. Adding a routine to directly compute this matrix directly would eliminate these issues. This would also eliminate the need for #246.

timryanb:
Adding geometric stiffness implementation for `TACSShellElement`.
Implementation currently works, tested on both tri and quad elements.
Should significantly improve accuracy of buckling derivatives in real and complex mode (no more finite differencing over a finite difference).

Closes #246,  Closes #294


sean-engelstad (approved):
Code is looking good! Eigenvalues and buckling modes match for unstiffened panels with old complex-step buckling solution and the new exact G matrix buckling solution.

## Testing panel buckling verification cases from my `ml_buckling` smdogroup repo

### Case 1 - 3 x 1 aluminum unstiffened plate under axial loading,

mode 1- new (TACS eigval = 3.8507, CPT closed-form eigval = 4.000)
image

mode 1- complex-step (TACS eigval = 3.851)
Screenshot from 2025-01-29 19-55-30


mode 2 - new (TACS eigval = 4.2018)
Screenshot from 2025-01-29 19-59-35


mode 2 - complex-step (TACS eigval = 4.202 )
Screenshot from 2025-01-29 19-55-23


## Case 2 - Unstiffened Metal Panel under Shear Load
The closed-form solution is not exact here, so I decided to compare modes and eigenvalues against previous complex-step case.

mode 1 - new (TACS eigval = 4.4387)
image

mode 1- complex-step (TACS eigval = 4.4387)
Screenshot from 2025-01-29 19-17-00


mode 3 - new (TACS eigval = 5.4138 )
Screenshot from 2025-01-29 19-17-47


mode 3 - complex-step (TACS eigval = 5.4138)
Screenshot from 2025-01-29 19-51-03

PR #355, opened 2025-01-29
Add support for discontinuous tecplot output

A-CGray:
1. Modifies f5totec to take a `--discontinuous` command line argument, in which case nodes are duplicated so that element data is plotted as it is written out by TACS, rather than averaging values at nodes which leads to unphysical looking results for shells and beams etc. Some examples are shown below.
2. Ran clang format on f5totec and f5tovtk
3. Add f5totec and f5tovtk to clang format checks run on PRs

image

image

image

### DIFF
--- a/.github/workflows/clang-format-check.yml
+++ b/.github/workflows/clang-format-check.yml
@@ -1,6 +1,6 @@
 name: clang-format Check
 
-on: 
+on:
   push:
     branches: [ master ]
   pull_request:
@@ -14,6 +14,7 @@ jobs:
         path:
           - 'src'
           - 'examples'
+          - 'extern'
     steps:
     - uses: actions/checkout@v4
     - name: Run clang-format style check for C/C++ programs.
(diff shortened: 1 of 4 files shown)


timryanb:
For the discontinuous mode, it looks like some of the element values are getting assigned to the wrong elements. I managed to reproduce the issue on a tacs [example](https://github.com/smdogroup/tacs/blob/master/examples/beam/ibeam_example.py) case. The error appears to have to do with the number of procs used to run the model when the f5 file is generated. Heres what the discrete output looks like when running the example in serial (i.e. `mpirun -np 1 python ibeam_example.py`):
image (7)

And heres what it looks like when running on 8 procs (i.e. `mpirun -np 8 python ibeam_example.py`):
image (8)

A-CGray:
I can reproduce this effect (below with `mpiexec -n 4`) so not a clang vs gcc thing

image

PR #356, opened 2025-01-29
Fixing remaining check_partials attributes in mphys wrappers

timryanb:
Apparently I missed a few `check_partials` attributes in PR #348

### DIFF
--- a/tacs/mphys/coupling.py
+++ b/tacs/mphys/coupling.py
@@ -25,7 +25,7 @@ class TacsCouplingGroup(om.Group):
 
     def setup(self):
         self.fea_assembler = self.options["fea_assembler"]
-        self.check_partials = self.options["check_partials"]
+        self.under_check_partials = self.options["check_partials"]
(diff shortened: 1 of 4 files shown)


A-CGray (approved):

tacs/mphys/coupling.py
```diff
@@ -25,7 +25,7 @@ def initialize(self):
 
     def setup(self):
         self.fea_assembler = self.options["fea_assembler"]
-        self.check_partials = self.options["check_partials"]
+        self.under_check_partials = self.options["check_partials"]
```
A-CGray, line 28: For consistency, I think we should change the name of the option to `under_check_partials` too

A-CGray (changes requested):

tacs/mphys/coupling.py
timryanb: Since that would be backwards incompatible I'd prefer to avoid this. But if you feel strongly I can reconsider

A-CGray (approved):

PR #358, opened 2025-02-27
updating openmdao tests to work with upcoming version release

timryanb:
This Pr should fix issues that will arise when OpenMDAO [PR #3465](https://github.com/OpenMDAO/OpenMDAO/pull/3465) is merged in

### DIFF
--- a/tests/integration_tests/openmdao_analysis_base_test.py
+++ b/tests/integration_tests/openmdao_analysis_base_test.py
@@ -158,8 +158,8 @@ class OpenMDAOTestCase:
                             rev_error = clean_data[component][in_out_tuple][
                                 error_type
                             ].reverse
-                            clean_data[component][in_out_tuple][
-                                error_type
-                            ] = ErrorTuple(0.0, rev_error, 0.0)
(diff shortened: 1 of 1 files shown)


tests/integration_tests/openmdao_analysis_base_test.py
```diff
@@ -111,6 +111,8 @@ def test_partials(self):
                 method=self.fd_method,
                 form=self.fd_form,
                 step=self.dh,
+                # TODO: For now we have to skip the solver partials until we implement a proper Jacobian transpose routine
+                excludes=["*.coupling.solver"],
```
A-CGray, line 115: Does this skip all partial checks for the solver? Are there not some we can still test (e.g residual partials w.r.t DVs and node coodinates)

timryanb: Unfortunately, OpenMDAO doesn't allow for that level of granularity at this time. You can only exclude on a component level

A-CGray: Damn, guess we need to get back to #336 at some point soon then

PR #361, opened 2025-03-05
Clean up compiler warnings

timryanb:
Minor edit to replace all `sprintf` functions with safer `snprintf` in C++ code.


examples/four_bar_mechanism/four_bar_mechanism.cpp
gjkennedy, line 1: should this be "filename" and not "file_name"?

gjkennedy (approved):

PR #364, opened 2025-03-24
removing numpy 2.0 upperbound

timryanb:
I've tested numpy 2.0 and it works with pyNastran 1.5.0-dev, so I'm removing the version upperbound from the tacs requirements


timryanb:
> Nice! Can we make a patch version release so that we have a version to point to when we try and update our docker images to numpy 2?

Sure, though as heads up you'll need to install the devel version of pyNastran from source. The pypi version doesn't support numpy 2.0 yet

PR #369, opened 2025-05-06
fixing indexing bug in f5totec

timryanb:
Fixing bug in f5totec where array was being incorrectly indexed by component number rather than by basic element number.  Could cause issues with post-processing mixed element f5 models


A-CGray:
Is there an example I can test this on?

timryanb:
> Is there an example I can test this on?

Yes, I'll forward you Chung Lee's email with the model @A-CGray

A-CGray (approved):

PR #371, opened 2025-05-08
Improved Buckling GP model with Buckling + RQ kernel functions

sean-engelstad:
* closed-form model extrapolation study => buckling + RQ kernel functions improve the extrapolation R^2 to 0.99 or 99% vs. best case 70% with other kernel functions. This is an improvement from previous kernel functions which didn't extrapolate well
* Lighter designs due to more accurate (less conservative) buckling predictions in extrapolation region of mixed SS+CL finite element dataset

[8_GP_closed_form_extrap.drawio.pdf](https://github.com/user-attachments/files/20105153/8_GP_closed_form_extrap.drawio.pdf)


sean-engelstad:
it's weird, this test that failed passed on my local, will look at it again later

timryanb (approved):

PR #375, opened 2025-05-30
bumping mpi4py min version to 4.0.3

timryanb:
The PyPI wheels for mpi4py 3.1.5 aren't available for Python 3.13+, which causes the cython build process to fail when installing the python interface for TACS. To be forward compatible with newer versions of Python we need to bump our minimum mpi4py version to 4.0.

Hoping to get a few people to try compiling TACS with this branch before we finalize this.


tacs/mpi-compat.h
```diff
@@ -11,4 +11,9 @@ typedef void *PyMPI_MPI_Message;
 #define MPI_Message PyMPI_MPI_Message
 #endif
 
+#if (MPI_VERSION < 4) && !defined(PyMPI_HAVE_MPI_Session)
+typedef void *PyMPI_MPI_Session;
+#define MPI_Session PyMPI_MPI_Session
+#endif
```
A-CGray, line 17: What does this do?

A-CGray:
On my machine with python 3.10.9, mpi4py 4.1.0:

```
OK

Passed:  557
Failed:  0
Skipped: 6
```

tacs/mpi-compat.h
timryanb: This ensures compatibility for mpi versions older than MPI 4.0 standard (openmpi 3.0 for example), since MPI_Session doesn't exist for those

PR #377, opened 2025-06-16
bumping min python version to 3.10

timryanb:
Python 3.9 is slated for EOL in October: https://devguide.python.org/versions/. We need to prepare to deprecate it and move the min version up to 3.10. Some dependencies like OpenMDAO have already made this [move](https://github.com/OpenMDAO/OpenMDAO/pull/3538).


A-CGray:
@timryanb, I don't know much about macOS, but seems like the build failure here might be related to [this issue](https://github.com/pymc-devs/pytensor/issues/1342), both have the same `ld: -lto_library library filename must be 'libLTO.dylib'` error message

timryanb:
Thanks this looks helpful. I'll take a look at this on Monday. > @timryanb, I don't know much about macOS, but seems like the build failure here might be related to [this issue](https://github.com/pymc-devs/pytensor/issues/1342), both have the same `ld: -lto_library library filename must be 'libLTO.dylib'` error message

timryanb:
Should be fixed now

A-CGray (approved):

PR #382, opened 2025-09-16
Fixing potential bug in pymeshloader

timryanb:
Potential bug in pymeshloader when bdf file features property cards with no elements assigned. 
The error is caused when a property ID dict is being looped over while entries are being deleted.
Updated a test to verify that the fix works.


A-CGray:
What was the error that this caused before you fixed it?

A-CGray (approved):

PR #385, opened 2025-09-22
Adding PBARL to pytacs

timryanb:
- Introduced a new test case for a 6-noded beam model using PBARL properties, validating various functions including KSDisplacement, KSFailure, StructuralMass, and Compliance.
- Added a corresponding BDF input file for the beam model to facilitate automatic initialization and testing.

### DIFF
--- a/tacs/pytacs.py
+++ b/tacs/pytacs.py
@@ -32,12 +32,12 @@ from functools import wraps
 import numpy as np
 import pyNastran.bdf as pn
 
-import tacs.TACS
 import tacs.constitutive
 import tacs.constraints
 import tacs.elements
 import tacs.functions
 import tacs.problems
+import tacs.TACS
 from tacs.pymeshloader import pyMeshLoader
 from tacs.utilities import BaseUI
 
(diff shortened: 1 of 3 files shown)


A-CGray (changes requested):
Sorry to be a pain, but can you add tests for the other two con types that the PBARL can be interpreted as? Not just the tube.

A-CGray:
> Yeah, I plan to do that next. I need to get access to nastran to double check the ordering for the dimensions in the bar beam case

The dimension ordering seems to be given [here](https://help.autodesk.com/view/NINCAD/2025/ENU/?guid=GUID-4B1B8CD3-80A5-49D0-8366-913D880C6B32)

timryanb:
> > Yeah, I plan to do that next. I need to get access to nastran to double check the ordering for the dimensions in the bar beam case
> 
> The dimension ordering seems to be given [here](https://help.autodesk.com/view/NINCAD/2025/ENU/?guid=GUID-4B1B8CD3-80A5-49D0-8366-913D880C6B32)

Thanks, I know the ordering of the dimensions relative to the cross-section definition. What I want to double check is that TACS and Nastran are using the same reference axis definition for beams. I believe this is the case, but beam reference axis definition is often times confusing which is why I'm hoping to have a nastran run to double check

timryanb:
> Sorry to be a pain, but can you add tests for the other two con types that the PBARL can be interpreted as? Not just the tube.

More tests added

tests/integration_tests/test_beam_bend_coupling.py
```diff
@@ -170,3 +173,14 @@ def setup_tacs_problems(self, comm):
...
+    # We have to skip these tests in complex mode because the beam
+    # element uses complex step to approximate the Jacobian and this
+    # leads to issues with complex stepping the sensitivities.
+    @unittest.skipIf(TACS_IS_COMPLEX, "Skipping test_total_dv_sensitivities")
+    def test_total_dv_sensitivities(self):
+        super().test_total_dv_sensitivities()
+
+    @unittest.skipIf(TACS_IS_COMPLEX, "Skipping test_total_xpt_sensitivities")
+    def test_total_xpt_sensitivities(self):
+        super().test_total_xpt_sensitivities()
```
A-CGray, line 186: How come this wasn't an issue previously?

timryanb: I'm not sure. There error I was seeing the tes was really finicky. I had a beam test that worked routinely and the modified one of the dvs to the 16th digit (ie. `t=0.01000000000000001`) and the complex sensitivity tests suddenly fail by a significant margin. To be safe I don't think we should be using cs with these elements since they are already using cs under the hood.

A-CGray (approved):

PR #389, opened 2025-09-25
Add postprocessing documentation for TACS solutions

timryanb:
- Introduced a new section on postprocessing capabilities and troubleshooting in the documentation.
- Added details on creating and converting f5 files, including output flags and visualization tips.
- Updated the index to include the new postprocessing section.

### DIFF
--- a/docs/source/postprocessing.rst
+++ b/docs/source/postprocessing.rst
@@ -1,0 +1,304 @@
+Postprocessing TACS Solutions
+=============================
+
+TACS provides comprehensive postprocessing capabilities through its native f5 file format and conversion utilities. This section explains how to generate f5 files from TACS analyses and convert them to standard visualization formats.
+
+F5 File Format
+--------------
+
+TACS uses the f5 file format (based on HDF5) to store solution data in a parallel, binary format. F5 files contain both continuous (nodal) and element-wise data, making them suitable for detailed postprocessing and visualization.
+
+Creating F5 Files
+~~~~~~~~~~~~~~~~~
+
+To create an f5 file from a TACS analysis, use the ``TACSToFH5`` class:
+
+.. code-block:: cpp
+
+   #include "TACSToFH5.h"
+   
+   // Create an TACSToFH5 object for writing output to files
+   ElementType etype = TACS_BEAM_OR_SHELL_ELEMENT;
+   int write_flag = (TACS_OUTPUT_CONNECTIVITY | TACS_OUTPUT_NODES |
+                     TACS_OUTPUT_DISPLACEMENTS | TACS_OUTPUT_STRAINS |
+                     TACS_OUTPUT_STRESSES | TACS_OUTPUT_EXTRAS);
+   TACSToFH5 *f5 = new TACSToFH5(assembler, etype, write_flag);
+   f5->incref();
+   f5->writeToFile("solution.f5");
+   f5->decref();
+
+Output Flags
+~~~~~~~~~~~~
+
+The following output flags control what data is written to the f5 file:
+
+- ``TACS_OUTPUT_CONNECTIVITY``: Element connectivity information
+- ``TACS_OUTPUT_NODES``: Nodal coordinates (X, Y, Z)
+- ``TACS_OUTPUT_DISPLACEMENTS``: Nodal displacements and rotations
+- ``TACS_OUTPUT_STRAINS``: Element strains
+- ``TACS_OUTPUT_STRESSES``: Element stresses
+- ``TACS_OUTPUT_EXTRAS``: Additional quantities (failure indices, design variables)
+- ``TACS_OUTPUT_LOADS``: Applied loads
+- ``TACS_OUTPUT_COORDINATE_FRAME``: Element coordinate frames
+
+Converting F5 Files
+-------------------
+
+TACS provides two utilities for converting f5 files to standard visualization formats:
+
+f5totec: Convert to Tecplot Format
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The ``f5totec`` utility converts f5 files to Tecplot format (.plt files):
+
+.. code-block:: bash
+
+   # Basic conversion
+   f5totec solution.f5
+   
+   # Convert with strands (for time-dependent data)
+   f5totec --use_strands solution.f5
+
+This creates a ``solution.plt`` file that can be opened in Tecplot.
+
+f5tovtk: Convert to VTK Format
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The ``f5tovtk`` utility converts f5 files to VTK format (.vtk files) for use with ParaView:
+
+.. code-block:: bash
+
+   f5tovtk solution.f5
+
+This creates a ``solution.vtk`` file that can be opened in ParaView.
+
+Output Variables by Element Type
+--------------------------------
+
+The following tables describe the output variables available for each element type in TACS.
+
+Beam/Shell Elements (TACS_BEAM_OR_SHELL_ELEMENT)
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. list-table:: Beam/Shell Element Output Variables
+   :widths: 20 20 60
+   :header-rows: 1
+
+   * - Category
+     - Variable
+     - Description
+   * - Displacements
+     - u, v, w
+     - Translational displacements
+   * - 
+     - rotx, roty, rotz
+     - Rotational displacements
+   * - Strains
+     - ex0, ey0, exy0
+     - Membrane strains
+   * - 
+     - ex1, ey1, exy1
+     - Bending strains
+   * - 
+     - eyz0, exz0
+     - Transverse shear strains
+   * - 
+     - erot
+     - Rotational strain
+   * - Stresses
+     - sx0, sy0, sxy0
+     - Membrane stress resultants
+   * - 
+     - sx1, sy1, sxy1
+     - Bending stress resultants
+   * - 
+     - syz0, sxz0
+     - Transverse shear stress resultants
+   * - 
+     - srot
+     - Rotational stress
+   * - Extras
+     - failure0-failure6
+     - Failure indices for different failure criteria
+   * - 
+     - dv1-dv7
+     - Design variables
+   * - Loads
+     - fx, fy, fz
+     - Applied forces
+   * - 
+     - mx, my, mz
+     - Applied moments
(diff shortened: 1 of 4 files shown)


docs/source/postprocessing.rst
```diff
@@ -0,0 +1,471 @@
...
+     - Applied moments
+   * - Coordinate Frame
+     - t0x, t0y, t0z
+     - First tangent vector components
+   * - 
+     - t1x, t1y, t1z
+     - Second tangent vector components
+   * - 
+     - t2x, t2y, t2z
+     - Normal vector components
```
A-CGray, line 198: I would call these 1,2,and 3 directions rather than tangent and normal

docs/source/postprocessing.rst
```diff
@@ -0,0 +1,471 @@
...
+   * - Loads
+     - Q
+     - Applied heat source
+
+Visualization Tips
+------------------
+
+1. **Element-wise vs. Nodal Data**: F5 files contain both element-wise and nodal data. The conversion utilities automatically perform nodal averaging for element-wise quantities.
+
+2. **Higher-order Elements**: Higher-order elements are automatically converted to basic element types for visualization (e.g., quadratic triangles become linear triangles).
```
A-CGray, line 330:
```suggestion
2. **Higher-order Elements**: Higher-order elements are split into multiple lower order element for visualization (e.g., each quadratic triangle becomes 3 linear triangles).
```

docs/source/postprocessing.rst
```diff
@@ -0,0 +1,471 @@
...
+2. **Create new variables** for deformed coordinates:
+   - Go to ``Data > Alter > Specify Equations``
+   - Create new variables:
+     - ``{XDEF} = {X} + {u}`` (deformed X coordinate)
+     - ``{YDEF} = {Y} + {v}`` (deformed Y coordinate)  
+     - ``{ZDEF} = {Z} + {w}`` (deformed Z coordinate)
+3. **Create the deformed plot**:
+   - Go to ``Plot > 3D``
+   - Set ``X``, ``Y``, ``Z`` to ``XDEF``, ``YDEF``, ``ZDEF``
+   - Choose appropriate surface rendering (``Surface``, ``Mesh``, or ``Contour``)
```
A-CGray, line 358: This is irrelevant to plotting the deformed surface

docs/source/postprocessing.rst
```diff
@@ -0,0 +1,471 @@
...
+   - Create variable: ``{DISP_MAG} = sqrt({u}*{u} + {v}*{v} + {w}*{w})``
+   - Set ``Contour`` to ``DISP_MAG`` to color-code by displacement magnitude
+
+**Example Tecplot equations:**
+::
+
+   {XDEF} = {X} + {u}
+   {YDEF} = {Y} + {v}
+   {ZDEF} = {Z} + {w}
+   {DISP_MAG} = sqrt({u}*{u} + {v}*{v} + {w}*{w})
```
A-CGray, line 369: Why not just include this in the steps above?

docs/source/postprocessing.rst
```diff
@@ -0,0 +1,471 @@
...
+
+   // 1. Create f5 file from TACS analysis
+   ElementType etype = TACS_BEAM_OR_SHELL_ELEMENT;
+   int write_flag = (TACS_OUTPUT_CONNECTIVITY | TACS_OUTPUT_NODES |
+                     TACS_OUTPUT_DISPLACEMENTS | TACS_OUTPUT_STRAINS |
+                     TACS_OUTPUT_STRESSES | TACS_OUTPUT_EXTRAS);
+   TACSToFH5 *f5 = new TACSToFH5(assembler, etype, write_flag);
+   f5->writeToFile("wing_analysis.f5");
+   f5->decref();
+
```
A-CGray, line 450: A python example would be more useful

docs/source/troubleshooting.rst
```diff
@@ -0,0 +1,211 @@
...
+
+**ImportError: No module named 'tacs'**
+
+*Cause*: TACS is not installed or not in the Python path.
+
+*Solutions*:
+
+- Ensure conda environment is activated: ``conda activate tacs``
+- Reinstall TACS: ``mamba install -c conda-forge -c smdogroup tacs``
+- For source installation, ensure Python interface is built: ``pip install -e .[all]``
```
A-CGray, line 17:
```suggestion
- For source installation, ensure Python interface is built: ``make interface``
```

docs/source/troubleshooting.rst
```diff
@@ -0,0 +1,211 @@
...
+
+*Cause*: Poor conditioning, incorrect boundary conditions, or excessive loads.
+
+*Solutions*:
+
+- Check boundary conditions are properly applied
+- Verify loads are reasonable for the structure
+- Ensure material properties are correct
+- Try reducing load magnitude
+- Check for rigid body modes
```
A-CGray, line 56: This advice is pretty generic. Is it really useful?

docs/source/troubleshooting.rst
```diff
@@ -0,0 +1,211 @@
...
+         else:
+             raise ValueError(f"Unexpected element type: {descript}")
+
+**Memory errors with large models**
+
+*Cause*: Insufficient memory for analysis or postprocessing.
+
+*Solutions*:
+
+- Consider using more powerful hardware
```
A-CGray, line 81: Again, is the advice "use more powerful hardware if you run out of memory" really useful? It would be much more useful to explain what output you'll see if you ran out of memory and if there are any options you can set in TACS to reduce memory usage

docs/source/troubleshooting.rst
```diff
@@ -0,0 +1,211 @@
...
+**Slow analysis performance**
+
+*Cause*: Suboptimal settings or hardware limitations.
+
+*Solutions*:
+
+- Use appropriate number of MPI processes
+- Enable optimized BLAS/LAPACK (Intel MKL)
+- Use SuiteSparse for better sparse matrix performance
+- Check system resources: ``htop``
```
A-CGray, line 122: These bullet points are not particularly useful unless they explain how to do these things, or link to somewhere that does. For example, how do I know what an "appropriate number of MPI processes" is? How do I enable TACS to use MKL? What does it mean to "Use SuiteSparse for better sparse matrix performance" and how do I do it? What should I be checking `htop` for?

docs/source/troubleshooting.rst
```diff
@@ -0,0 +1,211 @@
...
+**HPC System Issues**
+
+*Cause*: Module conflicts or resource limits.
+
+*Solutions*:
+
+- Use system-provided modules when available
+- Check job resource limits
+- Use appropriate compiler flags
+- Contact system administrators
```
A-CGray, line 157: As above, these solutions are all pretty vague

docs/source/troubleshooting.rst
```diff
@@ -0,0 +1,211 @@
...
+   from tacs import functions, constitutive, elements, pyTACS
+   print("All modules imported successfully")
+
+Getting Help
+------------
+
+**Community Resources:**
+
+- `GitHub Issues <https://github.com/smdogroup/tacs/issues>`_: Report bugs and request features
+- `TACS Documentation <https://smdogroup.github.io/tacs/>`_: Comprehensive documentation
```
A-CGray, line 183: Do we need to link to the documentation in a page that's already in the documentation?

A-CGray (changes requested):

docs/source/postprocessing.rst
```diff
@@ -0,0 +1,471 @@
...
+.. tip::
+   The ``ElementType`` should match the type of elements in your model. Common types include:
+   - ``TACS_BEAM_OR_SHELL_ELEMENT`` for shell and beam elements
+   - ``TACS_SOLID_ELEMENT`` for 3D solid elements
+   - ``TACS_PLANE_STRESS_ELEMENT`` for 2D plane stress elements
+
+Output Flags
+~~~~~~~~~~~~
+
+The following output flags control what data is written to the f5 file. You can combine multiple flags using the bitwise OR operator (``|``):
```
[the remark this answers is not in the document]
timryanb: I agree we should add a section on how to do the equivalent through pyTACS, but I think its still relevant to keep the C++ approach

docs/source/postprocessing.rst
```diff
@@ -0,0 +1,471 @@
...
+   # Convert with strands (for time-dependent data)
+   f5totec --use_strands solution.f5
+   
+   # Specify output filename
+   f5totec solution.f5 -o wing_analysis.plt
+
+This creates a ``solution.plt`` file that can be opened in Tecplot.
+
+.. tip::
+   The ``--use_strands`` option is particularly useful for transient analyses or optimization histories where you want to animate the results over time.
```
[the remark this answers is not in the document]
timryanb: Yeah, I just took a look at the source for this and it doesn't really make sense. I'll remove this from the docs, but we should probably figure out how to fix this or remove it from f5totec

docs/source/postprocessing.rst
```diff
@@ -0,0 +1,471 @@
...
+
+This creates a ``solution.vtk`` file that can be opened in ParaView.
+
+.. note::
+   Both utilities automatically handle the conversion of element-wise data to nodal data through averaging, making the results suitable for smooth visualization.
+
+**Troubleshooting Conversion Issues:**
+- Ensure the f5 file was generated successfully and contains the expected data
+- Check that the conversion utilities are compiled and accessible in your PATH
+- For large files, conversion may take several minutes - this is normal
```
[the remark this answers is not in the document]
timryanb: So the only case that I could think of for this would be the topology optimization cases @gjkennedy and his students were working with a while back. I think in those cases the models could be in the 100's of millions of elements. I'm not sure how long f5totec would take convert files of that size. But I agree, for most users this probably won't be relevant

docs/source/troubleshooting.rst
timryanb: I actually, do think its useful to remin d users to check bcs when their solve fails to converge. I'm always suprised by how many people will set up and try to run models without ever sanity checking them.

docs/source/troubleshooting.rst
timryanb: Agreed, let's remove.

docs/source/postprocessing.rst
timryanb: I assume you mean the 3rd bullet? Could you clarify @A-CGray

timryanb: Nevermind, gotcha. I'll update this to match Tecplots actual UI

A-CGray (changes requested):
Happy to merge after this final comment

A-CGray (approved):

PR #390, opened 2025-09-30
Expand Nastran beam card support

A-CGray:
## ToDo:

- [x] Support extraction of non-structural mass, offsets for centre of mass, centroid, and center of mass
  - [x] Implement extraction from PBEAM cards
  - [x] Implement extraction from PBEAML cards
  - [x] Expand cython interfaces of beam constitutive classes to support these arguments 
- [x] Add tests:
  - [x] PBEAM Cards
  - [x] PBEAML with "BAR" section
  - [x] PBEAML with "TUBE" section
  - [x] PBEAML with other section


tacs/pytacs.py
```diff
@@ -1157,10 +1157,65 @@ def elemCallBack(
...
+                        I12 = I12[0]
+                        J = J[0]
+                    else:
+                        xStations = propInfo.xxb
+                        area = np.trapz(area, xStations)
+                        I1 = np.trapz(I1, xStations)
+                        I2 = np.trapz(I2, xStations)
+                        I12 = np.trapz(I12, xStations)
+                        J = np.trapz(J, xStations)
+
```
timryanb, line 1178: Could we make this its own if block separate from PBAR to simplify things

tacs/pytacs.py
```diff
@@ -1157,10 +1157,65 @@ def elemCallBack(
...
+                sectionProps = {}
+                if sectionType == "BAR":
+                    sectionProps["w"] = propInfo.dim[:, 0]
+                    sectionProps["t"] = propInfo.dim[:, 1]
+                    conType = tacs.constitutive.IsoRectangleBeamConstitutive
+                elif propInfo.Type == "TUBE":
+                    r1 = propInfo.dim[:, 0]
+                    r0 = propInfo.dim[:, 1]
+                    sectionProps["d_inner"] = 2 * r0
+                    sectionProps["t_wall"] = r1 - r0
```
timryanb, line 1194: It looks like these keys are used later as arguments for the constitutive class. If that's the case, these should be "d" and "t" for IsoTubeBeamConstitutive

A-CGray:
@timryanb for some reason the ubuntu tests are running with numpy 1.26 instead of numpy>=2, which is the cause of a lot of these test failures. Any idea how that might be happening?

<https://github.com/smdogroup/tacs/actions/runs/30282659436/job/90032661774?pr=390#step:8:45>

timryanb:
> @timryanb for some reason the ubuntu tests are running with numpy 1.26 instead of numpy>=2, which is the cause of a lot of these test failures. Any idea how that might be happening?
> 
> https://github.com/smdogroup/tacs/actions/runs/30282659436/job/90032661774?pr=390#step:8:45

Should be fixed now @A-CGray. Had to push an updated pyspline conda package w/ numpy 2+ support
https://github.com/smdogroup/tacs/actions/runs/30282659436/job/90341276218?pr=390#step:8:45

PR #393, opened 2025-10-03
Adding interface for MACH framework

timryanb:
Adds wrapper for working with MDOLab's MACH framework

TODO:
- [x] Finish up docs
- [x] Add example
- [x] Add integration test


timryanb:
Still need to clean up docs

tacs/mach/struct_problem.py
```diff
@@ -0,0 +1,1487 @@
...
+        # Save a copy of the rhs vector holding the full loads
+        self.temp0.copyValues(rhs)
+        # Replace vector with only external loads
+        rhs.copyValues(self._Fext)
+        # Write external loads to bdf
+        self.staticProblem.writeLoadToBDF(fileName, loadCaseID=0)
+        # Reset rhs back to full loads
+        rhs.copyValues(self.temp0)
+
+    def readExternalForceFile(self, fileName):
```
eirikurj, line 1473: I did not see it, but I think it would be great to add a test this and `writeExternalForceFile` to make sure these are consistent.

tests/integration_tests/test_mphys_struct_hemisphere_nonlinear.py
```diff
@@ -58,6 +58,7 @@ def setup_problem(self, dtype):
         """
 
         # Overwrite default tolerances
+        self.atol = 5e-4
```
eirikurj: Dont remember why this was added as part of this work. Just want to make sure that this is correct.

tacs/mach/struct_problem.py
```diff
@@ -0,0 +1,1487 @@
...
+        -------
+        dict
+            Dictionary containing transpose matrix-vector product results.
+        """
+        # # The old way (for testing purposes)
+        prodDV = self.FEAAssembler.createDesignVec(asBVec=True)
+        for f in evalFuncs:
+            f_mangled = self.name + "_%s" % f
+            self.staticProblem.addDVSens(
+                [evalFuncs], [prodDV], scale=vecT[f_mangled][0]
```
eirikurj: Dont think we are testing this routine anywhere, but I think `[evalFuncs]` should be `[f]` in the loop. Same below.

tests/integration_tests/mach_struct_problem_base_test.py
```diff
@@ -0,0 +1,314 @@
...
+                                        rtol=self.rtol,
+                                        atol=self.atol,
+                                    )
+
+        def test_total_dvgeo_sensitivities(self):
+            """
+            Test total geometric design variable sensitivity through adjoint against fd/cs
+            """
+            # Skip this check if no dvs were added to model
+            total_geo_dvs = self.dv_geo.getNDV()
```
eirikurj: I think `self.dv_geo` can be `None` so this might fail if tests do not set DVGeo (should we add one?). Maybe add a check here and just return if `None`?

docs/source/interfaces.rst
```diff
@@ -2,13 +2,14 @@ Interfaces
...
+There are currently several supported approaches for interfacing with this functionality in Python: direct, pyTACS, MPhys, MACH, and caps2tacs.
 These methods are documented below.
 
 .. toctree::
   :maxdepth: 1
 
   core/TACS
   pytacs/pytacs
-  mphys/mphys
+  mphys/mphy
```
eirikurj: looks like an `s` got dropped

examples/beam/shape_opt.py
```diff
@@ -0,0 +1,155 @@
...
+# Now we create the structural optimization problem:
+optProb = Optimization("Mass min", structObj)
+optProb.addObj("tip_shear_mass")
+structProb.addVariablesPyOpt(optProb)
+DVGeo.addVariablesPyOpt(optProb)
+optProb.addCon("tip_shear_ks_vmfailure", upper=1.0)
+
+optProb.printSparsity()
+
+optOptions = {
```
eirikurj: Did you mean to use SNOPT for this example or is this some leftover stuff? Maybe add an option to the example to support both optimizers?

docs/source/mach/mach.rst
```diff
@@ -0,0 +1,71 @@
...
+- **Adjoint Capabilities**: The class manages all adjoint state vectors needed for coupled sensitivity analysis, including :attr:`~tacs.mach.struct_problem.StructProblem.phi`, :attr:`~tacs.mach.struct_problem.StructProblem.pLoad`, :attr:`~tacs.mach.struct_problem.StructProblem.dSdu`, and :attr:`~tacs.mach.struct_problem.StructProblem.adjRHS`.
+- **Geometric Design Variables**: When a ``DVGeometry`` object is provided (via :meth:`~tacs.mach.struct_problem.StructProblem.setDVGeo`), node coordinates are automatically updated before each analysis call and geometric sensitivities are folded in by :meth:`~tacs.mach.struct_problem.StructProblem.evalFunctionsSens`.
+- **pyoptsparse Integration**: :meth:`~tacs.mach.struct_problem.StructProblem.addVariablesPyOpt` and :meth:`~tacs.mach.struct_problem.StructProblem.addConstraintsPyOpt` register structural design variables and constraints directly with a ``pyoptsparse`` optimization problem.
+- **Force File I/O**: External coupling forces can be saved to and loaded from BDF-format files for post-processing or restarting aerostructural analyses.
+
+Adjoint Vector Properties
+--------------------------
```
eirikurj: I am getting build warning since the underline is too long

docs/source/mach/mach.rst
```diff
@@ -0,0 +1,71 @@
...
+- **Adjoint Capabilities**: The class manages all adjoint state vectors needed for coupled sensitivity analysis, including :attr:`~tacs.mach.struct_problem.StructProblem.phi`, :attr:`~tacs.mach.struct_problem.StructProblem.pLoad`, :attr:`~tacs.mach.struct_problem.StructProblem.dSdu`, and :attr:`~tacs.mach.struct_problem.StructProblem.adjRHS`.
+- **Geometric Design Variables**: When a ``DVGeometry`` object is provided (via :meth:`~tacs.mach.struct_problem.StructProblem.setDVGeo`), node coordinates are automatically updated before each analysis call and geometric sensitivities are folded in by :meth:`~tacs.mach.struct_problem.StructProblem.evalFunctionsSens`.
+- **pyoptsparse Integration**: :meth:`~tacs.mach.struct_problem.StructProblem.addVariablesPyOpt` and :meth:`~tacs.mach.struct_problem.StructProblem.addConstraintsPyOpt` register structural design variables and constraints directly with a ``pyoptsparse`` optimization problem.
+- **Force File I/O**: External coupling forces can be saved to and loaded from BDF-format files for post-processing or restarting aerostructural analyses.
+
+Adjoint Vector Properties
```
eirikurj: Nitpicking, but would it make sense to split `Fext` out since its not strictly an adjoint vector? 
Alternatively, jut rename `Managed State Vectors` or something.

tacs/mach/struct_problem.py
```diff
@@ -0,0 +1,1487 @@
...
+
+        # Compute actual final FEA Norm
+        res.zeroEntries()
+        self.staticProblem.addJacVecProduct(self._phi, res, transpose=True)
+        res.axpy(-1.0, self._adjRHS)  # Add the RHS
+        self.staticProblem.finalNorm = np.real(res.norm())
+
+    def getdRdXptPhi(self, objectives):
+        """
+        Get the result of :math:`[dR/dX_{nodes}]^T \phi` for each of the objectives in the objective list.
```
eirikurj: I am getting warnings when running and building docs here and below
```
struct_problem.py:1304: SyntaxWarning: invalid escape sequence '\p'
struct_problem.py:1326: SyntaxWarning: invalid escape sequence '\p'
```
we should change the string to raw or add a another backslash

docs/source/conf.py
```diff
@@ -212,6 +212,13 @@
         None,
     ),
     "mphys": ("https://openmdao.github.io/mphys/", None),
+    "pygeo": ("https://mdolab-pygeo.readthedocs-hosted.com/en/latest/", None),
+    "pyoptsparse": (
+        "https://mdolab-pyoptsparse.readthedocs-hosted.com/en/latest/",
+        None,
+    ),
+    "mach": ("https://mdolab-mach-aero.readthedocs-hosted.com/", None),
```
eirikurj: I am getting a failure for this link, its probably just missing the `en/latest` like the other ones

docs/source/conf.py
```diff
@@ -212,6 +212,13 @@
...
+    "adflow": ("https://mdolab-adflow.readthedocs-hosted.com/en/latest/", None),
 }
 
 autodoc_mock_imports = ["mphys"]
```
eirikurj: do we want to add `pygeo` and `pyoptsparse` here as well? Otherwise, building the docs will fail if these are not installed.

eirikurj (commented):
Looks good. As discussed, I focused on the docs (and docstrings) and skimmed the rest. Only a few very minor comments.

tests/integration_tests/test_mphys_struct_hemisphere_nonlinear.py
timryanb: I assume this was due to a failing test. This test in particular has been finicky in the past

tacs/mach/struct_problem.py
timryanb: Added this to a new integration test

eirikurj (approved):

eirikurj:
Looks good! I did not merge in case @A-CGray wants to review also. If not feel free to merge.

tests/integration_tests/test_mphys_struct_hemisphere_nonlinear.py
A-CGray: We shouldn't need this any more, https://github.com/smdogroup/tacs/pull/438 changed this test to skip the total derivative checks in real mode

examples/beam/shape_opt.py
A-CGray: Yeah we should definitely update this to use one of the built-in pyoptsparse optimisers, likely slsqp or psqp

examples/beam_shape_opt/shape_opt.py
A-CGray, line 1: This `beam` directory now has 4 completely separate examples in it (`analysis.py`, `beam_opt.py`, `ibeam_example.py`, and now `shape_opt.py`). I think we should move them each to their own directory.

examples/beam/shape_opt.py
A-CGray: Oh nvm, the example does actually use slsqp, let's just delete these SNOPT settings

A-CGray:
I converted the beam shape optimization example documentation to grab code blocks directly from the example script using `literalinclude`, rather than us manually copying them into the rst file. We should do this for the other docs examples in a separate PR

examples/beam_shape_opt/shape_opt.py
A-CGray: Also, when I run this example I get a lot of `TACSRBE3: Error, points are colinear. Make sure to specify sufficient independent DOF's to react all moments.`. Even if this is not a fatal error I think we should fix it.

tacs/mach/struct_problem.py
```diff
@@ -0,0 +1,1483 @@
...
+        ----------
+        value : numpy.ndarray or tacs.TACS.Vec
+            Adjoint RHS vector to set.
+        """
+        self._adjRHS[:] = value
+
+    @property
+    def dIdu(self):
+        """
+        Get the derivative of objective function with respect to state variables.
```
A-CGray: This is a bit picky but can we remove all the uses of the word "objective" when we're really just talking about any function? This is done in many docstrings.

tacs/mach/struct_problem.py
```diff
@@ -0,0 +1,1483 @@
...
+        dvVal = self.comm.bcast(dIdXdv, root=0)
+
+        return dvVal
+
+    def getdIduTransProd(self, vecT, evalFuncs=None):
+        """
+        Perform the transpose matvec of the getdIduProd function and
+        return a TACS vector. The result is stored locally.
+
+        We assume that structProblem is set and that evalFuncs is valid.
```
A-CGray: I presume this is copied over from old code, the `structProblem` being set is not a relevant concern any more right? We're in the structProblem here

tacs/mach/struct_problem.py
```diff
@@ -0,0 +1,1483 @@
...
+        rhs = self.staticProblem.rhs
+        # Save a copy of the rhs vector holding the full loads
+        self.temp0.copyValues(rhs)
+        # Replace vector with only external loads
+        rhs.copyValues(self._Fext)
+        # Write external loads to bdf
+        self.staticProblem.writeLoadToBDF(fileName, loadCaseID=0)
+        # Reset rhs back to full loads
+        rhs.copyValues(self.temp0)
+
```
A-CGray: This needs to be updated based on the changes you made in https://github.com/smdogroup/tacs/pull/447 right?

tacs/mach/struct_problem.py
```diff
@@ -0,0 +1,1483 @@
...
+            only used from an external solver.
+        number : int or None
+            Use the user supplied number to index solution. Again, only
+            typically used from an external solver
+        """
+        self.staticProblem.writeSolution(outputDir, baseName, number)
+
+    def writeExternalForceFile(self, outputDir=None, baseName=None, number=None):
+        """
+        This function writes the external loads to a file.
```
A-CGray: Is this method supposed to write out all the external forces (i.e including inertial loads etc), or just the loads stored in `self._Fext`? Should clarify this in the docstring

tests/integration_tests/test_mach_beam_dvgeo.py
```diff
@@ -0,0 +1,117 @@
...
+        staticProb = FEAAssembler.createStaticProblem("tip_shear")
+        # Add TACS Functions
+        staticProb.addFunction("mass", functions.StructuralMass)
+        staticProb.addFunction(
+            "ks_vmfailure", functions.KSFailure, safetyFactor=1.0, ksWeight=ksweight
+        )
+        # Add forces to static problem
+        staticProb.addLoadToNodes(
+            206, [0.0, V, 0.0, 0.0, 0.0, 0.0], nastranOrdering=True
+        )
```
A-CGray: I think we should add a few different kinds of loads here, including through `setExternalForce`, so that `test_mach_beam_load_file.py` is a more thorough test.

A-CGray (changes requested):

tests/integration_tests/test_mach_beam_dvgeo.py
timryanb: Are you saying we should add these loads to test_mach_beam_load_file or test_mach_beam_dvgeo @A-CGray ?

examples/beam_shape_opt/shape_opt.py
timryanb: @A-CGray Is this showing up towards the end of the optimization? If so, it's likely because the RBE3 at the tip of the beam is being shrunk down to a singular point by the FFD. So that's triggering a singular tolerance check built into TACSRBE3

tacs/mach/struct_problem.py
timryanb: It should only write out aero loads. Docstring updated for clarity

examples/beam_shape_opt/shape_opt.py
A-CGray: Yeah only in the last few iterations

tests/integration_tests/test_mach_beam_dvgeo.py
A-CGray: Not sure exactly how you want to arrange, but to test the the reading and writing of loads files is consistent we should do the following:

1. Setup a problem with multiple loads added through pyTACS (e.g intertial loads) plus nodal forces set through `setExternalForce`
2. Call `writeExternalForceFile` on that problem
3. Setup an identical problem with all the same loads, except instead of calling `setExternalForce` we call `readExternalForceFile`
4. Check that the two problems produce the same results

timryanb: 3. Unfortunately won't work because writeExternalLoadFile only writes aero loads out (e.g. loads saved in self.Fext)

timryanb:
@A-CGray is there anything left to do on this PR?

A-CGray (approved):

eirikurj (approved):

PR #398, opened 2025-10-17
Modifying f5totec to properly write out element data

timryanb:
Modified f5totec to specify element data as "cell-centered" when writing out data with TECIO. Should fix issues with element data "bleeding" across element boundaries.
Before PR:
image
After PR:
image

### DIFF
--- a/extern/f5totec/f5totec.cpp
+++ b/extern/f5totec/f5totec.cpp
@@ -332,8 +332,7 @@ int main( int argc, char * argv[] ){
 
     int *reduced_points = new int[ num_points ];
     int *reduced_conn = new int[ basic_conn_size ];
-    int* value_location = NULL;
-    value_location = new int[ num_variables + num_evariables ];
+    int* value_location = new int[ num_variables + num_evariables ];
     for ( int j = 0; j < num_variables; j++ ){
       value_location[j] = 1; // 1 = nodal
     }


timryanb:
This PR also seems to fix the issue we were seeing in PR #355 where the element data writing would be messed up for cases run on multiple procs with a mixture of element types. 

Here's the I beam example output run on 10 procs, which is the case we were noticing issues on in PR #355
image

extern/f5totec/f5totec.cpp
```diff
@@ -330,14 +322,24 @@ int main( int argc, char * argv[] ){
...
+      belem_global_ptr += ntypes;
     }
 
     int num_comp = loader->getNumComponents();
 
     int *reduced_points = new int[ num_points ];
     int *reduced_conn = new int[ basic_conn_size ];
-    float *reduced_float_data = NULL;
-    reduced_float_data = new float[ num_points ];
+    int* value_location = NULL;
```
A-CGray: Is there a reason for having this line as opposed to just doing `int *value_location = new int[ num_variables + num_evariables ];`?

extern/f5totec/f5totec.cpp
```diff
@@ -226,31 +228,19 @@ int main( int argc, char * argv[] ){
...
+    // For each element, average the nodal data
+    float *avg_edata = new float[ num_elements * num_evariables ];
+    memset(avg_edata, 0, num_elements * num_evariables*sizeof(float));
+    for ( int i = 0; i < num_elements; i++ ){
+      int nnodes = ptr[i+1] - ptr[i];
+      for ( int j = 0; j < num_evariables; j++ ){
+        for ( int k = ptr[i]; k < ptr[i+1]; k++ ){
+          avg_edata[num_evariables*i + j] += edata[num_evariables*k + j];
+        }
+        avg_edata[num_evariables*i + j] /= nnodes;
```
A-CGray: So we're not averaging the nodal values anymore and tecio can handle the discontinuous nodal values? Nice.

PR #412, opened 2026-02-05
Deprecation warning for `addTransposeJacVecProduct` method in `StaticProblem` 

timryanb:
PR #404 removed `addTansposeJacVecProduct` from `StaticProblem` class. This is technically a breaking change, so I've added it back with a deprecation warning to be removed in version 3.12.0


A-CGray (approved):
Good catch, I should have added this in my PR but the MPhys wrapper was the only place I could see it being called.

PR #433, opened 2026-04-07
Add `ruff` workflow and enforce stricter linting rules

eirikurj:
## Description
This PR replaces `black` with a `ruff` formatting and linting CI workflow. 

The defaults for formatting should be the same as the ones found in `black`. For example, the default line length is 88 characters. 

The `ruff` linting rules are added to the `pyproject.toml` file along with some common exceptions. The [linting exceptions](https://docs.astral.sh/ruff/rules/) are minimal but necessary in my opinion, but there are some there that we might want to work on getting removed in the future.

Note, currently there are tests and examples that will fail the linting and need attention. In some cases its not clear what the author's intent was, and in others more extensive refactoring could be done (e.g. `setup_tacs_vecs` is largly the same in many tests and could be shared across many).

Closes #429


A-CGray:
@eirikurj I fixed the outstanding linting issues but there are a lot of changes so I opened a PR (https://github.com/mdolab/tacs/pull/6) to this branch so you can review

A-CGray (approved):

PR #436, opened 2026-04-10
fixing bug in xpt sens for StiffenerLengthConstraint

timryanb:
The node coordinate sensitivities in SparseLengthConstraint.evalConSens were computed using raw node coordinates instead of the displacement vector dX = X_end1 - X_end0. The correct derivative of L = |dX| w.r.t. an end node position requires dX/L, not the node's absolute position. This produced wrong gradients whenever the stiffener was not centered at the origin.

The existing gradient check in test_rectangle_beam_buckling.py failed to catch this because the test mesh (beam_model_translated.bdf) had one of the end nodes located at (0, 0, 0) and so the error canceled out. Updated the example mesh so that issue won't be missed in future.

### DIFF
--- a/tacs/constraints/stiffener_length.py
+++ b/tacs/constraints/stiffener_length.py
@@ -278,7 +278,7 @@ class StiffenerLengthConstraint(TACSConstraint):
         Examples
         --------
         >>> funcs = {}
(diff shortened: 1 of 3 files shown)


A-CGray (approved):
Sorry, I have OCD w.r.t repeated code

eirikurj (approved):

PR #438, opened 2026-04-17
Decreasing default shell drilling regularization to better match Nastran

timryanb:
Previous NASTRAN validation studies have shown that TACS is typically overly stiff relative to equivalent NASTRAN models. This excess stiffness is largely attributable to the `Drilling Regularization` parameter used by shell constitutive classes. Tuning this value is non-trivial: too high and the solution diverges from ideal behavior; too low and the linear system becomes numerically unstable.

This PR reduces the default drilling regularization value for shell elements from 10.0 to 0.1. This value has proven to be a good balance between accuracy and numerical stability in practice. To validate the change, I swept through a range of values using the I-beam [example](https://github.com/smdogroup/tacs/blob/master/examples/beam/ibeam_example.py) model (shown below) and compared each solution against the deflection predicted by NASTRAN.

image
image


timryanb:
This change will require a version bump due to it being a braking change

A-CGray (approved):

tests/integration_tests/test_shell_hemisphere_nonlinear.py
```diff
@@ -16,34 +16,34 @@
...
 
 # nonlinear quadratic
 hemisphereProbRefFuncs = {
-    "RadialForces_Compliance": 20.59867522072626,
-    "RadialForces_KSFailure": 2.3731860388536483,
-    "RadialForces_MaxYDisp": 0.4948883544992733,
-    "RadialForces_MaxZDisp": 0.4093283830600901,
+    "RadialForces_Compliance": 165.51140441529395,
+    "RadialForces_KSFailure": 6.827665576404779,
+    "RadialForces_MaxYDisp": 1.7454271859340547,
```
eirikurj, line 41: @A-CGray just wanted to confirm that you intended on removing the value `DRILLING_REGULARIZATION = 10.0`. The change in these values is huge. Do you know if these values compare better to NASTRAN?

src/constitutive/TACSShellConstitutive.cpp
```diff
@@ -56,7 +56,7 @@ void TACSShellConstitutive::extractTangentStiffness(
 /*
   Set the default drilling regularization value
 */
-double TACSShellConstitutive::DRILLING_REGULARIZATION = 10.0;
+double TACSShellConstitutive::DRILLING_REGULARIZATION = 0.1;
```
eirikurj, line 61: Dont see any documentation on how this was chosen. Should we maybe add a comment here (or elsewhere)?

eirikurj (commented):
Looks good, just some minor specific comments/clarifications.

Also, I think we should update the Cython layer to match the new default value, specifically [setDrillingRegularization](https://github.com/timryanb/tacs/blob/00b3d8d6e91c18845a1bad378576c400bae120d5/tacs/constitutive.pyx#L833) which seems to still use the default of 10.0. This might cause some unexpected behavior if one calls it without an argument.

This is secondary, but is it worth adding a test to check the value? I guess that requires a `getDrillingRegularization` to test the `setDrillingRegularization`, which may not have much use.

tests/integration_tests/test_shell_hemisphere_nonlinear.py
A-CGray: Yeah this was intentional. These values are much closer to published benchmark results for this problem (https://doi.org/10.1016/j.finel.2003.11.001) than before. I've yet to come across a case where reducing the drilling parameter to 0.1 gave less accurate results

timryanb:
> Looks good, just some minor specific comments/clarifications.
> 
> Also, I think we should update the Cython layer to match the new default value, specifically [setDrillingRegularization](https://github.com/timryanb/tacs/blob/00b3d8d6e91c18845a1bad378576c400bae120d5/tacs/constitutive.pyx#L833) which seems to still use the default of 10.0. This might cause some unexpected behavior if one calls it without an argument.
> 
> This is secondary, but is it worth adding a test to check the value? I guess that requires a `getDrillingRegularization` to test the `setDrillingRegularization`, which may not have much use.

Should be addressed now @eirikurj

eirikurj (approved):

eirikurj:
@timryanb looks good! Not sure if you wanted to wait for other feedback or not. If not, feel free to merge.

PR #439, opened 2026-04-17
Expose failure criterion and KS aggregation type as typed Python enums

timryanb:
## C++ changes

- Lifted `TACSKS{Failure,Temperature,Displacement}::KS…Type` and `TACSInducedFailure::InducedNormType` out of their classes into top-level `KSAggregationType` and `InducedNormType` enums in `TACSFunction.h`, prefixing the shared KS values with `KS_`.
- Renamed the per-class setters to a uniform `setKSAggregationType(KSAggregationType)`. `KS_DISCRETE_AVERAGE` is only valid for `KSFailure`; the other two reject it.
- Renamed `TACSOrthotropicPly::FailureCriterion` → `CompositeFailureCriterion` and replaced the five `setUseXxxCriterion()` methods with one `setFailureCriterion(...)`.

## Python / Cython changes

- New `CompositeFailureCriterion` and `KSAggregationType` `IntEnum`s, populated directly from the C++ enum values via Cython's rename trick (private `_C…` cast targets keep the Python class name free).
- Old boolean/string kwargs still work but emit `DeprecationWarning`; combining old + new raises `ValueError`. `KS_DISCRETE_AVERAGE` on `KSTemperature`/`KSDisplacement` raises `ValueError`.
- Fixes a pre-existing typo in `tacs/cpp_headers/functions.pxd` (`TACSKFailure` → `TACSKSFailure`, etc.) on the way through.

## Tests

- New `test_ks_aggregation_type_validation.py` — `KS_DISCRETE_AVERAGE` is rejected by `KSTemperature`/`KSDisplacement`, accepted by `KSFailure`.
- New `test_shell_comp_smeared_tsai_wu.py` — `KSFailure` and sensitivities under Tsai-Wu for a smeared composite layup.
- Existing examples and tests migrated to the new kwargs.

## Compatibility

Python surface is backward-compatible with a deprecation cycle. Direct C++ consumers must update to the renamed enum/setter names.


tacs/constitutive.pyx
```diff
@@ -613,7 +615,8 @@ cdef class OrthotropicPly:
     cdef TACSOrthotropicPly *ptr
     cdef MaterialProperties props
     def __cinit__(self, TacsScalar ply_thickness, MaterialProperties props,
-                  max_strain_criterion=False, Cuntze_criterion_UD=False, Cuntze_criterion_Woven=False):
+                  max_strain_criterion=False, tsai_wu_criterion=False, 
+                  Cuntze_criterion_UD=False, Cuntze_criterion_Woven=False):
```
A-CGray: Can we expose the [`FAILURE_CRITERION` enum](https://github.com/smdogroup/tacs/blob/298b2a8288b89cd1a0166dda6cf5b30d271c5caf/src/constitutive/TACSMaterialProperties.h#L177-L183) through cython and then just have a single `failure_criterion` input argument here that takes an instance of that enum? Right now a user could pass `True` for multiple failure criteria and it's not clear which one would be used.

A-CGray (changes requested):

tacs/constitutive.pyx
timryanb: That's the approach I wanted to take initially, but didn't want to break backwards compatibility. I might do that anyways and raise a deprecation warning for the old kwargs

A-CGray: If we do that, could we also change the name of the enum to `COMPOSITE_FAILURE_CRITERION` as well? I don't think that will break anything in the public API because it's currently only used privately right?

tacs/constitutive.pyx
```diff
@@ -588,46 +589,70 @@ cdef class MAT2MaterialProperties(MaterialProperties):
...
+        elif failure_criterion == FC.TSAI_WU_MODIFIED:
+            self.ptr.setUseModifiedTsaiWuCriterion()
+        elif failure_criterion == FC.CUNTZE_UD:
             self.ptr.setUseCuntzeCriterion_UD()
-        elif Cuntze_criterion_Woven:
+        elif failure_criterion == FC.CUNTZE_WOVEN:
             self.ptr.setUseCuntzeCriterion_Woven()
         else:
-            self.ptr.setUseModifiedTsaiWuCriterion()
+            raise ValueError(f'Unknown failure_criterion: {failure_criterion!r}')
```
A-CGray: Not a fan of the fact that we're diverging the python and C++ API for OrthoPly here. Can we update the C++ class to use the same interface we're defining here? (i.e get rid of the `setUse*Criterion` methods and replace them with a single `setFailureCriterion` method that takes an instance of `TACSOrthotropicPly::FailureCriterion` as input)? That would allow us to get rid of all this if else logic, and maybe also this new `_OrthotropicPly` class?

A-CGray (changes requested):

A-CGray:
@timryanb I opened a PR (https://github.com/timryanb/tacs/pull/75) to this branch that address my comments, I'll let you decide whether to merge them since they result in some more substantial changes than this PR originally did.

tacs/constitutive.pyx
A-CGray: This is resolved by https://github.com/timryanb/tacs/pull/75

A-CGray (approved):
Good from my side, but we should definitely have @eirikurj review too since I contributed to these changes.

tacs/functions.pyx
```diff
@@ -226,15 +242,13 @@ cdef class KSTemperature(Function):
     Args:
         assembler (Assembler): TACS Assembler object that will evaluating this function.
         ksWeight (float, optional): The ks weight used in the calculation (keyword argument). Defaults to 80.0.
-        ftype (str, optional): The type of KS aggregation to be used (keyword argument).
-            Accepted inputs are: 'discrete', 'continuous', 'pnorm-discrete', and 'pnorm-continuous'.
-            Case-insensitive, defaults to 'continuous'.
+        ks_aggregation_type (functions.KSAggregationType, optional): The type of KS aggregation to be used.
```
eirikurj: Should we not use camelCase for the arguments and API (all other arguments here and elsewhere use that)?

tacs/functions.pyx
```diff
@@ -393,24 +453,54 @@ cdef class KSDisplacement(Function):
...
+            ks_aggregation_type (functions.KSAggregationType or str): The aggregation type.
+                String values are deprecated; use ``functions.KSAggregationType`` instead.
+                ``DISCRETE_AVERAGE`` is not supported for this function.
+
+        Raises:
+            ValueError: If ``ks_aggregation_type`` is ``KSAggregationType.KS_DISCRETE_AVERAGE``.
+        """
+        if isinstance(ks_aggregation_type, str):
+            warnings.warn(
+                "Passing a string to setKSAggregationType is deprecated. "
```
eirikurj: Do you want to add a version for when the deprecation warning is removed? This is not specific to this PR and regularly comes up, and maybe we should discuss it more generally, but we should have some policy and a way of communicating how long these will stay and be supported in the code. Otherwise, these tend to stay indefinitely.

tacs/functions.pyx
```diff
@@ -393,24 +453,54 @@ cdef class KSDisplacement(Function):
...
+                "Use a functions.KSAggregationType enum value instead.",
+                DeprecationWarning,
```
eirikurj: I dont see any tests, but should we not test and catch for the `DeprecationWarning`?

tacs/constitutive.pyx
```diff
@@ -588,46 +589,76 @@ cdef class MAT2MaterialProperties(MaterialProperties):
...
+            warnings.warn(
+                f"Boolean failure criterion kwargs ({', '.join(k for k, _ in _deprecated_used)}) are deprecated. "
+                "Use 'failure_criterion=constitutive.CompositeFailureCriterion.<VALUE>' instead.",
+                DeprecationWarning,
+                stacklevel=2,
+            )
+            if len(_deprecated_used) > 1:
+                raise ValueError('Only one failure criterion can be specified.')
+            if failure_criterion is not None:
+                raise ValueError(
```
eirikurj, line 692: Didnt find any tests for this `ValueError` when specifying both. Worth testing these?

src/functions/TACSKSTemperature.cpp
```diff
@@ -48,7 +48,13 @@ const char *TACSKSTemperature::funcName = "TACSKSTemperature";
 /*
   Set the KS aggregation type
 */
-void TACSKSTemperature::setKSTemperatureType(enum KSTemperatureType type) {
+void TACSKSTemperature::setKSAggregationType(KSAggregationType type) {
+  if (type == KS_DISCRETE_AVERAGE) {
```
eirikurj, line 52: To take advantage of the validation here, I think we should probably call this function from the constructor, in case anyone is using c++ directly.

tacs/functions.pyx
```diff
@@ -249,31 +263,59 @@ cdef class KSTemperature(Function):
...
+                "The 'ftype' string kwarg is deprecated. "
+                "Use 'ks_aggregation_type=functions.KSAggregationType.<VALUE>' instead.",
+                DeprecationWarning,
+                stacklevel=2,
+            )
+            if 'ks_aggregation_type' in kwargs:
+                raise ValueError(
+                    "Cannot specify both 'ks_aggregation_type' and deprecated 'ftype' kwarg."
+                )
+            self.setKSAggregationType(kwargs['ftype'])
```
eirikurj: will this call potentially trigger a second deprecation warning since the `ftype` is still a string? Maybe cast to enum before to avoid triggering in `setKSAggregationType`?

tacs/functions.pyx
```diff
@@ -249,31 +263,59 @@ cdef class KSTemperature(Function):
...
+                DeprecationWarning,
+                stacklevel=2,
+            )
+            _str_map = {
+                'discrete': KSAggregationType.KS_DISCRETE,
+                'continuous': KSAggregationType.KS_CONTINUOUS,
+                'pnorm-discrete': KSAggregationType.PNORM_DISCRETE,
+                'pnorm-continuous': KSAggregationType.PNORM_CONTINUOUS,
+            }
+            ks_aggregation_type = _str_map[ks_aggregation_type.lower()]
```
eirikurj: I did not see any tests, but if someone passes a sting that is not in the map this will probably fail with a `KeyError`. Do you want to add and catch a `KeyError`?

eirikurj (commented):
Overall looks good, just a few minor comments mostly related to testing or consistency.

tacs/functions.pyx
A-CGray: Yes you're right, fixed now

tacs/functions.pyx
A-CGray: I have removed eveything string-related out of `setKSAggregationType`. Now when the user passes in a string, the deprecation warning is raised once and the string converted to an enum instance before `setKSAggregationType` is called.

src/functions/TACSKSTemperature.cpp
A-CGray: Done

tacs/constitutive.pyx
A-CGray: This is now covered by `test_unknown_ftype_raises_value_error`

tacs/functions.pyx
A-CGray: Done

tacs/functions.pyx
A-CGray: There are now tests to check this is raised

tacs/functions.pyx
A-CGray: I added this. Also did a bit of research on managing deprecations, there are a couple of packages (https://pypi.org/project/Deprecated/, https://pypi.org/project/pyDeprecate/) we could use both for implementing deprecation warnings and for automatically detecting deprecations that are due to be removed. There are also multiple existing deprecations in TACS that are past their removal date. We should fix these and discuss a strategy in a separate issue/PR.

timryanb: Created new issue [here](https://github.com/smdogroup/tacs/issues/450).

tacs/constitutive.pyx
```diff
@@ -588,46 +589,78 @@ cdef class MAT2MaterialProperties(MaterialProperties):
...
+    The interaction coefficient for the Tsai-Wu failure criterion is set
+    to zero by default. If a value of C, the failure stress under
+    combined in-plane loading, is supplied, the interaction coefficient
+    is determined. Be careful - the value can easily fall outside
+    acceptable bounds - these are tested during initialization.
+
+    Args:
+        ply_thickness (float or complex): The ply thickness.
+        props (MaterialProperties): The ply material property.
+        failure_criterion (constitutive.CompositeFailureCriterion): The failure criterion to use.
```
eirikurj: Missed this before, but `failure_criterion` is also `snake_case`. In fact, the arguments here are all `snake_case` (with mixed capitalization), although most of them are labeled deprecated. Do we want to use this opportunity to change `ply_thickness` to `plyThickness` etc. This may depend on my other comment if we are introducing a breaking change.

tacs/functions.pyx
```diff
@@ -322,30 +361,51 @@ cdef class KSFailure(Function):
...
-        self.setKSFailureType(ftype)
+            ksAggregationType = kwargs.get('ksAggregationType', KSAggregationType.KS_CONTINUOUS)
+        self.setKSAggregationType(ksAggregationType)
 
-        return
+    def setKSAggregationType(self, ksAggregationType):
+        """
+        Set the type of KS aggregation.
 
-    def setKSFailureType(self, ftype='discrete'):
```
eirikurj, line 332: Just thinking here in terms of deprecation warnings and backwards compatibility. This method `setKSFailureType` (and others e.g., `setKSDisplacementType` below) has been removed in favor of a new interface.

Strictly speaking, this is a change in API and thus not a backward-compatible change in the python layer. Do we want to keep it backwards compatible by keeping these functions and adding a small wrapper and a warning, or are we happy with breaking backwards compatibility? This will for sure create issues in lots of scripts.

eirikurj (commented):
The changes look good; just a few follow-up comments or clarification related to backwards compatibility.

tacs/constitutive.pyx
A-CGray: Done, `failure_criterion` is now `failureCriterion`. `ply_thickness` is now `plyThickness` but `ply_thickness` is still supported with a deprecation warning.

tacs/functions.pyx
A-CGray: Done, added back `setKSFailureType` etc methods that raise deprecation warnings

tacs/constitutive.pyx
```diff
@@ -642,23 +714,14 @@ cdef class OrthotropicPly:
         """
         return self.props
 
-    def setUseMaxStrainCriterion(self):
```
eirikurj, line 645: Missed this before, but similar to my earlier comment about the API, these three functions `setUseMaxStrainCriterion, setUseTsaiWuCriterion, setUseModifiedTsaiWuCriterion` are gone. We probably need to wrap these as well if we don't want to break the API. 
I think this is it, but please check.

tests/constitutive_tests/test_orthotropic_ply_deprecation.py
```diff
@@ -0,0 +1,161 @@
+"""Tests for OrthotropicPly failure-criterion API: new enum kwarg, deprecated
+boolean kwargs, and the validation errors raised for bad combinations.
+Also covers the plyThickness/ply_thickness kwarg rename."""
+
+import unittest
+import warnings
+from importlib.metadata import version
+
+from packaging.version import Version
```
eirikurj: The `packaging` is a new package and we might want to add it as a `testing` dependency

eirikurj (commented):
Few minor comments

tests/constitutive_tests/test_orthotropic_ply_deprecation.py
A-CGray: Done

A-CGray:
I've implemented a `fails_at_version` decorator that you can add to any test to make it fail at or above a specific TACS version. Going forward we should use this decorator on any tests that test a deprecation. The idea being that when we go to update the version of TACS in a PR, any deprecation tests that we were supposed to remove in that version will fail if we've forgotten to remove them.

tacs/constitutive.pyx
A-CGray: Done. I checked for others and the only other method removed from the python API is `KSTemperature.setLoadFactor`, but that method doesn't actually exist in the C++ layer so that's more of a bugfix than a deprecation.

tacs/_testing.py
eirikurj, line 1: Is there a reason to place the file in the `tacs` dir? Should we not place this file directly under `tests` as this is only meant for tests? Somewhat confusing also, as this file will be package with the `tacs`.

A-CGray: I did so that it's importable from tacs. If we put this in the `tests` dir then we'll need to start using `sys.path.append` and path-based imports, which I don't like

eirikurj (approved):

PR #445, opened 2026-05-01
Add IsoTubeBeam buckling and CompositeTubeBeam constitutive with topology 

markleader:
## Summary

This PR adds new beam constitutive capabilities for tube-based sizing/topology optimization and fills in a missing derivative path for point-mass frequency sensitivities.

## What Changed

- Added `TACSCompositeTubeBeamConstitutive` for hollow circular composite tube beams.
- Extended `TACSIsoTubeBeamConstitutive` with Euler column buckling support.
- Added topology-variable with SIMP scaling and a stiffness floor to the iso tube constitutive.
- Exposed the new and updated constitutives through the Python/Cython bindings.
- Added mass-matrix DV sensitivity inner products to `TACSMassElement` for point-mass contributions in modal/frequency problems.

### Composite tube beam constitutive

Introduces a new `TACSCompositeTubeBeamConstitutive` that:

- Computes smeared effective axial and shear properties from ply-level orthotropic properties and layup angles using CLT.
- Supports inner-diameter and wall-thickness design variables.
- Supports optional Euler buckling checks via `Lb`/`Kb`.
- Aggregates fiber compression, optional fiber tension, and Euler buckling failure modes with a KS smooth max.
- Provides analytic strain and design-variable sensitivities.

### Iso tube beam updates

Extends `TACSIsoTubeBeamConstitutive` to support:

- Euler buckling failure evaluation for tube members.
- Optional buckling-length design variables.
- Optional topology variables with SIMP-based density/stiffness scaling.
- A finite stiffness floor for near-void members to improve conditioning in eigenvalue solves.
- Updated failure, mass, stiffness, and derivative calculations consistent with the new buckling/topology behavior.

### Point mass sensitivity support

Updates `TACSMassElement` so point-mass constitutive design variables contribute through mass-matrix sensitivity inner products, which is needed for frequency/eigenvalue derivative calculations.

### Python bindings

Updates the Cython and C++ interface declarations to expose:

- New optional buckling/topology arguments on `IsoTubeBeamConstitutive`
- New `CompositeTubeBeamConstitutive`

## Motivation

These changes enable:

- Buckling-aware tube beam optimization workflows
- Composite tube beam modeling with analytic sensitivities
- Better-conditioned topology-aware modal optimization
- Correct point-mass derivative propagation in frequency problems


timryanb:
@markleader it looks like there are some errors with dividing a TacsScalar by an int. You should be able to fix this by re-casting the int var like shown in this [example](https://github.com/smdogroup/tacs/blob/master/src/functions/TACSKSFailure.cpp#L209).

It also looks like there's some minor clang format issues. Should be able to fix these files with: ``clang-format --style=Google -i filename.cpp``

markleader:
@timryanb I just fixed the integer-divides, and ran `clang-format` on all modified cpp files. I didn't run `clang-format` on the header files - let me know if those need it as well.

markleader:
@timryanb I fixed the formatting, but `style=Google` wasn't exactly right for me: I had to run `style=file` to fix the pointer-spacing, so I updated the `clang-format` instructions in the README, also.

src/constitutive/TACSIsoTubeBeamConstitutive.cpp
```diff
@@ -140,25 +226,33 @@ TacsScalar TACSIsoTubeBeamConstitutive::evalDensity(int elemIndex,
...
 }
 
 void TACSIsoTubeBeamConstitutive::addDensityDVSens(
     int elemIndex, TacsScalar scale, const double pt[], const TacsScalar X[],
     int dvLen, TacsScalar dfdx[]) {
   int index = 0;
   TacsScalar d0 = inner + wall;
+  TacsScalar d1 = inner;
   TacsScalar rho = props->getDensity();
+  TacsScalar rho_scale = (xDV >= 0) ? (eps_m + (1.0 - eps_m) * x_val) : 1.0;
```
timryanb: Can we move this code to a new subroutine? Something like TACSIsoTubeBeamConstitutive::getMaterialDensity. In that routine we can have the check for if the topology dv is active and if so return the scaled material density and otherwise return the regular density. You'll probably need a corresponding DV sens function as well.

src/constitutive/TACSIsoTubeBeamConstitutive.cpp
```diff
@@ -250,92 +372,199 @@ TacsScalar TACSIsoTubeBeamConstitutive::evalFailure(int elemIndex,
...
+  int dvNums[4];
+  dvNums[0] = innerDV;
+  dvNums[1] = wallDV;
+  dvNums[2] = buckleLengthDV;
+  dvNums[3] = xDV;
+  int index = 0;
 
-  // Compute the stress
-  props->evalStress3D(e0, s0);
+  TacsScalar x_relax = (xDV >= 0) ? x_val : 1.0;
```
timryanb: Lets make this a "get" method too

src/constitutive/TACSIsoTubeBeamConstitutive.cpp
```diff
@@ -1,9 +1,35 @@
 #include "TACSIsoTubeBeamConstitutive.h"
 
+namespace {
+// Finite SIMP stiffness floor for numerical regularization so near-void
+// members do not create near-singular stiffness matrices. This improves
+// conditioning for eigenvalue solves and parallel runs.
+inline TacsScalar simp_stiffness_scale(int xDV, TacsScalar x_val,
```
timryanb: It seems like all of these inputs are class member variables so we shouldn't need to pass them as inputs to this method

src/constitutive/TACSIsoTubeBeamConstitutive.cpp
```diff
@@ -170,6 +264,9 @@ void TACSIsoTubeBeamConstitutive::evalStress(int elemIndex, const double pt[],
   TacsScalar E, nu;
   props->getIsotropicProperties(&E, &nu);
 
+  TacsScalar stiffness_scale =
```
timryanb: I would suggest modifying this method to simply return the scaled modulus, like I suggested for the density. Something like `getYoungsModulus`

src/constitutive/TACSIsoTubeBeamConstitutive.cpp
```diff
@@ -313,12 +332,10 @@ void TACSIsoTubeBeamConstitutive::addStressDVSens(
...
     TacsScalar dfdx[]) {
-  TacsScalar E, nu;
-  props->getIsotropicProperties(&E, &nu);
-
-  TacsScalar stiffness_scale =
-      simp_stiffness_scale(xDV, x_val, p_penalty, k_floor);
-  TacsScalar Ep = E * stiffness_scale;
+  TacsScalar E0, nu;
+  props->getIsotropicProperties(&E0, &nu);
+  (void)E0;
```
timryanb: Could we save rho0, E0, and nu0 (unscaled values) as class members in the class init and then we only have to call props->getIsotropicProperties once.

timryanb (approved):

A-CGray:
@markleader there was a bug in the section property calculations for `TACSIsoTubeBeamConsitutive` that we just fixed in #455 , you may want to check whether you copied that bug into your new class. The errors are quite significant as it results in the cross-section area and MOI being off by a factor of 2 or more.

markleader:
> @markleader there was a bug in the section property calculations for `TACSIsoTubeBeamConsitutive` that we just fixed in #455 , you may want to check whether you copied that bug into your new class. The errors are quite significant as it results in the cross-section area and MOI being off by a factor of 2 or more.

I updated `TACSCompositeTubeBeam` to for the corrected wall-thickness convention, and updated the missing `const` declarations to match #455 .

src/constitutive/TACSCompositeTubeBeamConstitutive.cpp
```diff
@@ -0,0 +1,646 @@
+#include "TACSCompositeTubeBeamConstitutive.h"
+
+#include <math.h>
+#include <string.h>
+
+TACSCompositeTubeBeamConstitutive::TACSCompositeTubeBeamConstitutive(
+    TacsScalar E11_in, TacsScalar E22, TacsScalar G12, TacsScalar nu12,
+    TacsScalar rho_in, TacsScalar X_c_in, TacsScalar X_t_in,
```
A-CGray, line 8: In other constitutive classes, we don't pass individual material properties to the constructor, we pass an instance of `TACSMaterialProperties` or `TACSorthotropicPly`. Can you modify this class to use the same approach?

src/constitutive/TACSCompositeTubeBeamConstitutive.cpp
```diff
@@ -0,0 +1,646 @@
...
+  innerLb = inner_lb;
+  innerUb = inner_ub;
+  wallLb = wall_lb;
+  wallUb = wall_ub;
+
+  // Buckling parameters
+  buckleLength = buckle_length;
+  buckleLengthFactor = buckle_length_factor;
+
+  ks_weight = 100.0;
```
A-CGray, line 70: Can you add a method for changing this, e.g https://github.com/smdogroup/tacs/blob/2099a54c8bb6fd2556697d03a9e667cdb2bf75cf/src/constitutive/TACSBladeStiffenedShellConstitutive.h#L300-L306

src/constitutive/TACSCompositeTubeBeamConstitutive.cpp
```diff
@@ -0,0 +1,646 @@
...
+
+  // Fiber tension (optional)
+  if (TacsRealPart(X_t) > 0.0) {
+    fail_checks[count] = (E11 * eps1) / X_t;
+    if (TacsRealPart(fail_checks[count]) > TacsRealPart(max_fail)) {
+      max_fail = fail_checks[count];
+    }
+    count++;
+  }
+
```
A-CGray, line 437: If I'm understanding this code correctly, you're only computing fibre tension and compression for fibres aligned with the beam axis, but this class is modelling a beam made from a laminate with plies at arbitrary angles. To be more accurate, we should really be computing a failure value for each ply angle. 

If you switch using the `TACSorthotropicPly` as my other comment suggested, you can use the failure criteria already included in that class, they also handle the transformation of the strain based on the ply angle. Plus you would add the option to use other failure criteria (e.g Tsai-Wu) for free. 

Your code here would look pretty much identical to the code we use for computing material failure in the composite shell classes:

https://github.com/smdogroup/tacs/blob/2099a54c8bb6fd2556697d03a9e667cdb2bf75cf/src/constitutive/TACSBladeStiffenedShellConstitutive.cpp#L1457-L1487

src/constitutive/TACSCompositeTubeBeamConstitutive.cpp
```diff
@@ -0,0 +1,646 @@
...
+    TacsScalar const d0 = inner + 2.0 * wall;
+    TacsScalar const d1 = inner;
+    TacsScalar const A = M_PI * ((d0 * d0) - (d1 * d1)) / 4.0;
+    TacsScalar const Ia =
+        M_PI * ((d0 * d0 * d0 * d0) - (d1 * d1 * d1 * d1)) / 64.0;
+    TacsScalar const Leff = buckleLengthFactor * buckleLength;
+    TacsScalar const Nx = -A * e[0];
+    TacsScalar const Pcr = M_PI * M_PI * Ia / (Leff * Leff);
+    TacsScalar const buckle_ratio = Nx / Pcr;
+    fail_checks[count] = getTopologyValue() * buckle_ratio;
```
A-CGray, line 449: The buckling failure value is multiplied by `getTopologyValue()` but the material failure values are not. Is that deliberate?

src/constitutive/TACSCompositeTubeBeamConstitutive.cpp
```diff
@@ -0,0 +1,646 @@
...
+
+  eps_1 = e[0] + r0*(e[2] + e[3]) — axial strain at outer fibre.
+  Uses the same r0 convention as TACSIsoTubeBeamConstitutive: r0 = 0.5*inner +
+  wall.
+*/
+TacsScalar TACSCompositeTubeBeamConstitutive::evalFailure(
+    int elemIndex, const double pt[], const TacsScalar X[],
+    const TacsScalar e[]) {
+  TacsScalar const r0 = 0.5 * inner + wall;
+  TacsScalar const eps1 = e[0] + r0 * (e[2] + e[3]);
```
A-CGray, line 416: This is computing the strain at the OD based on the sum of the two bending strains, which doesn't seem correct to me. Shouldn't you be using `e[0] + r0 * sqrt(e[2]^2 + e[3]^2);` and `e[0] - r0 * sqrt(e[2]^2 + e[3]^2);` to get the maximum tensile and compressive strain?

A-CGray (changes requested):
Can't say I've looked through every line of code, but here are a few comments to get started.

Also, we need to add the following tests:
- In ``test_iso_tube_beam_constitutive.py` we should enable the new topology and buckling features you added so those derivatives are tested
- An equivalent of `test_iso_tube_beam_constitutive.py` for the new composite beam class
- Add a modal problem to `test_beam_point_mass.py` so that we test your new point mass DV sensitivity implementation

PR #447, opened 2026-05-14
Use externalForce vector for load norm and BDF load export in StaticProblem

timryanb:
- Replaces the manual K·u - res computation used to initialize the solver convergence norm with a call to getForces(), which cleanly separates the external and internal force components of the residual
- Uses externalForce.norm() as initNorm instead of the previously computed rhs norm, making the convergence criterion more physically meaningful (normalized against the applied load magnitude)
- Updates writeLoadToBDF to pull the load vector from self.externalForce rather than self.rhs, which more correctly represents the externally applied loads rather than the solver RHS (which could include solver-internal modifications)

### DIFF
--- a/tacs/problems/static.py
+++ b/tacs/problems/static.py
@@ -925,12 +925,13 @@ class StaticProblem(TACSProblem):
         # Get current residual
         self.getResidual(self.res, Fext=Fext)
 
-        # Get rhs vector
-        self.K.mult(self.u, self.rhs)
-        self.rhs.axpy(-1.0, self.res)
-
-        # Set initnorm as the norm of rhs
-        self.initNorm = np.real(self.rhs.norm())
+        # Compute the internal and external force components of the residual at the current point
+        self.getForces(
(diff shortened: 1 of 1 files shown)


tacs/problems/static.py
```diff
@@ -925,12 +925,13 @@ def _solveLinear(self, Fext=None):
...
-
-        # Set initnorm as the norm of rhs
-        self.initNorm = np.real(self.rhs.norm())
+        # Compute the internal and external force components of the residual at the current point
+        self.getForces(
+            externalForceVec=self.externalForce,
+            internalForceVec=self.internalForce,
+            Fext=Fext,
+        )
+        self.initNorm = np.real(self.externalForce.norm())
```
A-CGray: These lines are now identical in `_solveLinear` and `_solveNonlinear`, so we could move them to `solve` to avoid duplication

A-CGray (changes requested):

A-CGray (approved):

PR #453, opened 2026-06-02
Improve ESP/CAPS installation with error handling

timryanb:
CI will now skip ESP compilation if tar ball download timesout

### DIFF
--- a/.github/workflows/unit_tests.yml
+++ b/.github/workflows/unit_tests.yml
@@ -133,7 +133,7 @@ jobs:
           # Install ESP/CAPS prebuilt for shape derivatives, only in Real mode
            if [[ ${{ matrix.NAME }} == 'Real-Ubuntu' ]]; then
             cd $TACS_DIR/extern/
-            if wget --timeout 60 https://acdl.mit.edu/ESP/archive/ESP122Lin.tgz; then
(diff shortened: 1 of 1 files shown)


.github/workflows/unit_tests.yml
```diff
@@ -133,27 +133,30 @@ jobs:
...
-             ./config/makeEnv ${CASROOT}
-             source $ESP_ROOT/../ESPenv.sh
-             cd ./src/CAPS/aim
-             make
-             cd $TACS_DIR/extern/ESP122/
-             # remove all ESP/CAPS unit test files with recursive delete
-             find . -name "test*" -type f -delete
-             cd $TACS_DIR
-           fi
+            if wget https://acdl.mit.edu/ESP/archive/ESP122Lin.tgz; then
```
eirikurj: minimal, but the default `wget` timeout is 900 seconds, which is pretty long. I think this can probably be shortened to something like 60 seconds using the `--timeout=60` flag.

eirikurj: also, we should set `-tries=2` (default is 20) to avoid multiple tries each timing out

eirikurj (approved):

PR #457, opened 2026-06-12
Add shear correction factor argument to `TACSIsoShellConstitutive` constructor

A-CGray:
1. Adds `kcorr` as optional constructor argument
2. Set `kcorr` from bdf file inside `_elemCallBackFromBDF`
3. Update reference MatVec product values used in `test_matrix_operations.py` (necessary because the default pyNastran `kcorr` value is truncated to 0.833333 rather than full precision 5.0 / 6.0


src/constitutive/TACSIsoShellConstitutive.cpp
```diff
@@ -174,7 +174,7 @@ void TACSIsoShellConstitutive::evalStress(int elemIndex, const double pt[],
     }
 
     // Set the through-thickness shear stiffness
-    As[0] = As[2] = (5.0 / 6.0) * A[5];
+    As[0] = As[2] = (kcorr)*A[5];
```
eirikurj: I think we should remove the parenthesis around  `kcorr`. Same elsewhere.

src/constitutive/TACSIsoShellConstitutive.cpp
```diff
@@ -27,7 +27,7 @@ const char *TACSIsoShellConstitutive::constName = "TACSIsoShellConstitutive";
 */
 TACSIsoShellConstitutive::TACSIsoShellConstitutive(
     TACSMaterialProperties *props, TacsScalar _t, int _tNum, TacsScalar _tlb,
-    TacsScalar _tub, TacsScalar _tOffset) {
+    TacsScalar _tub, TacsScalar _kcorr, TacsScalar _tOffset) {
```
eirikurj: Just want to bring this up, but not sure we want to act on this or not since most users are using the python API. This is technically a breaking API change in the c++ layer and it should be a version bump. The argument order is different, and since we are using default values in the header, there will be no feedback and the user will now pass `_tOffset` as `_kcorr` and `_tOffset` will be set to the default.

eirikurj (changes requested):
See minor comments

tacs/constitutive.pyx
```diff
@@ -1129,12 +1131,14 @@ cdef class IsoShellConstitutive(ShellConstitutive):
             tlb = kwargs['tlb']
         if 'tub' in kwargs:
             tub = kwargs['tub']
+        if 'kcorr' in kwargs:
+            kcorr = kwargs['kcorr']
         if 'tOffset' in kwargs:
             tOff = kwargs['tOffset']
```
eirikurj: This was not changed here, but it seems that this variables is the only one that has a different naming strategy `tOffset` is `tOff`. Maybe unify?

src/constitutive/TACSIsoShellConstitutive.h
```diff
@@ -32,7 +32,8 @@ class TACSIsoShellConstitutive : public TACSShellConstitutive {
  public:
   TACSIsoShellConstitutive(TACSMaterialProperties *props, TacsScalar _t = 1.0,
                            int _tNum = -1, TacsScalar _tlb = 0.0,
-                           TacsScalar _tub = 1.0, TacsScalar _tOffset = 0.0);
+                           TacsScalar _tub = 1.0, TacsScalar _kcorr = 5.0 / 6.0,
```
eirikurj: I noticed that `_tub` does not have the same default value in the header and in the Cython layer (1.0 and 10.0, respectively). I think we should fix this to be consistent.

eirikurj (changes requested):
Couple of other minor comments (I was too quick to press submit before).

src/constitutive/TACSIsoShellConstitutive.cpp
A-CGray: I agree, done

src/constitutive/TACSIsoShellConstitutive.cpp
A-CGray: Yeah good catch, I moved kcorr to the end now

tacs/constitutive.pyx
A-CGray: Renamed tOff → tOffset to match the kwarg and the other constitutive classes. Also reordered the Python kwarg listing so it mirrors the C++ argument order.

src/constitutive/TACSIsoShellConstitutive.h
A-CGray: Fixed — set the header default to 10.0 to match the long-standing Python default (and the documented value).

eirikurj (approved):
Looks good

PR #462, opened 2026-07-07
 Add API for extracting component-wise design variable information

A-CGray:
After running an optimisation, it is common for the user to want to extract information about the design variable values for each component in their model. This is currently a P.I.T.A for several reasons:

1. The design variable vector is distributed across procs and doesn't contain any information about which values are associated with which components or design variable
2. Both the design variable vector and the constitutive class's `getDesignVars` only return information about active design variables, which may not fully define a model.

This PR adds new methods to pyTACS that produce user-friendly dictionaries containing the values and indices for all design variables associated with each component in a model, e.g:

```python
{"WING_SKIN.001": {"t": 0.0042}, "SPAR.001": {"panelThick": 0.003, "panelPlyFracs": array([0.5, 0.3, 0.2])}, ...}
```

To enable this, the following new methods have been added to the `TACSConstitutive` class and all of its child constitutive classes:

- `getNumDesignVarGroups`
- `getDesignVarGroupName`
- `getDesignVarGroupSize`
- `isDesignVarGroupScalar`
- `getDesignVarGroupValues`
- `getDesignVarGroupNums`

these methods return values for all of a constitutive class's potential design variables, whether they are active (dvNum > 0) or not.


timryanb:
I like the idea of this @A-CGray. One of my biggest frustrations with the TACS constraints like Adjacency constraints is having to provide the dv index that you want included in the constraint, since this changes based on what DVs are active in the component. Maybe we can add some of this API to the constraint classes to make them work consistently regardless of what dvs are active in the component

A-CGray:
> I like the idea of this @A-CGray. One of my biggest frustrations with the TACS constraints like Adjacency constraints is having to provide the dv index that you want included in the constraint, since this changes based on what DVs are active in the component. Maybe we can add some of this API to the constraint classes to make them work consistently regardless of what dvs are active in the component

Right, that's an excellent idea. Do you want to try and implement that in this PR? Or do it in a follow-up?

Also, I have another branch built off of this one that uses this new API in the output file writing so that our tecplot/paraview files include all DVs, and they have real names rather than just `dv0, dv1, ...`. Was going to open that once this is merged.

timryanb:
You don't have to implement the constraint refactoring in this PR. But maybe take a look through some of those constraint classes and think about how we can get rid of the dvIndex approach using this new API or if there's any other API we can add here that will help accomplish that.

timryanb:
I'm curious bout the tecplot PR. Will the dv names still work in cases w/ models using mixed element types (i.e. beams + shells)?

A-CGray:
> I'm curious bout the tecplot PR. Will the dv names still work in cases w/ models using mixed element types (i.e. beams + shells)?

Yeah there are fields for all the types of DV used across the model.

timryanb (approved):

A-CGray:
Let's hold off on merging this until I implement your suggestion we discussed in the dev meeting today @timryanb . I'll make `getComponentDesignVars` and `getComponentDesignVarNums` methods of the problem class to remove the ambiguity about which problem's DVs are being returned. I'll re-request your review when that's ready.

tacs/system.py
```diff
@@ -142,6 +144,188 @@ def getDesignVarRange(self):
...
+            seen.add(descript)
+        if len(duplicates) > 0:
+            raise self._TACSError(
+                "Component descriptions must be unique to use getComponentDesignVars/"
+                "getComponentDesignVarNums, but the following descriptions are shared by "
+                f"multiple components: {sorted(duplicates)}. Rename the property groups in "
+                "the BDF file so that every component has a unique description."
+            )
+
+    def _getComponentConstitutive(self, compID):
```
A-CGray, line 267: @timryanb I want to point out this limitation in the current implementation to you because it's easy to miss. These new methods currently only work if every element in any given TACS component uses the same constitutive object. This is usually what we do, but I could envision a scenario in which you have shell and beam elements under single component, in which case these new DV extraction methods wouldn't work.

I'm sure it wouldn't be too hard to implement support for that, it'd just be a question of how to represent it in the dictionary we return. Most obvious option would be to add another level of dictionary nesting so instead of the current 2 level structure:

```python
dvValue = dict["compDescript"]["dvGroupName"]
```

We'd have:

```python
dvValue = dict["compDescript"]["elemDescript"]["dvGroupName"]
```

A-CGray:
> I'll make `getComponentDesignVars` and `getComponentDesignVarNums` methods of the problem class to remove the ambiguity about which problem's DVs are being returned.

in the end I moved these methods to the `TACSSystem` class, so they're inherited by all problem and constraint classes and will return data reflecting that problem/constraint's design vector.

tacs/system.py
A-CGray: @timryanb

A-CGray:
Marked this as draft as https://github.com/smdogroup/tacs/pull/468 contains an alternative implementation that we might merge into this PR.

PR #463, opened 2026-07-10
Add design variable support for aux elements

timryanb:
Summary
This PR enables entries of the inertia (acceleration) vector passed to `addInertialLoad` and pressures passed to `addPressureToElements` and `addPressureLoadsToComponents` and tractions passed to `addTractionsToElements` and `addTractionsToComponents` to be treated as design variables in an optimization. This allows, for example, the magnitude of a gravity/pressure load to be a DV without modifying the structural stiffness model.

Example
```
# Register gravity magnitude as a global DV (returns its DV number)
gravDVNum = FEAAssembler.addGlobalDV("g", -9.81)

# FEAAssembler intialization stuff
...

# Create problem
problem = FEAAssembler.createStaticProblem("dummy")

# z-component is the DV; x and y are fixed
g = np.array([0.0, 0.0, -9.81])
problem.addInertialLoad(g, [-1, -1, gravDVNum])
```


tacs/mach/struct_problem.py
```diff
@@ -62,6 +64,19 @@ def __init__(
         loadFile : str, optional
             Filename of the (static) external load file. Should be
             generated from pyAerostructure.
+
+        promotedGlobalDVs : list of str or None, optional
+            Names of global DVs (registered via ``pyTACS.addGlobalDV``) that
+            should be promoted to the problem level.  Each promoted DV is
+            exposed to the optimizer as a separate scalar variable keyed
+            ``"{problemName}_{dvName}"`` so that multiple StructProblems can
```
A-CGray: Can you swap the order to `"{dvName}_{problemName}"` so that we match the convention of `AeroProblem` and pyGeo?

tacs/mach/struct_problem.py
```diff
@@ -62,6 +64,19 @@ def __init__(
         loadFile : str, optional
             Filename of the (static) external load file. Should be
             generated from pyAerostructure.
+
+        promotedGlobalDVs : list of str or None, optional
```
A-CGray, line 68: Do you think there's any way you could allow users to specify that a group of globalDVs should be combined into an array? For example, if I have global DVs controlling a bunch of mass elements, it would be convenient to have them all grouped together in a single `masses` DV array instead of separate `mass_0, mass_1,...` DVs.

timryanb: This might be better to accomplish at the pyTACS assembler level. We could add an optional kwarg to addGlobalDV that allows users to create vectors of dvs.

A-CGray: OK, let's save that for a separate PR then

timryanb: Already had Claude implement it @A-CGray, but can roll back the commit if we want to break into another PR

src/elements/shell/TACSBeamTraction.h
```diff
@@ -29,6 +32,11 @@ class TACSBeamTraction : public TACSElement {
         t[i] = _t[i];
       }
     }
+    if (_tDVNums) {
+      memcpy(tDVNums, _tDVNums, 3 * sizeof(int));
```
eirikurj: Not an issue, but I think we could use something more modern than `memcpy`. For example, we could use `std::copy_n` (here and elsewhere)

tacs/problems/static.py
```diff
@@ -800,9 +883,9 @@ def _updateAssemblerVars(self):
         the input variables associated with this problem
         """
 
+        self.assembler.setAuxElements(self.auxElems)
```
eirikurj: I see `setAuxElements` was reordered here to appear before the `setDesignVars`. I assume this is because of a stale aux elements then being used. Should this not happen in other problems as well? For example, the buckling one does not do this, and the transient one has a loop (over all elements, see my other comment also there). From what I can tell, the aux elements should be set before you call `setDesignVars`. Please check.

Aside from this would it be possible to come up with a mechanism that syncs this such that this is consistent across all problems? Maybe some type of sync decorator or something like that?

tacs/problems/buckling.py
```diff
@@ -552,10 +566,20 @@ def addTractionToElements(
...
-            self.auxElems, elemIDs, tractions, faceIndex, nastranOrdering
+            self.auxElems,
+            elemIDs,
+            tractions,
+            faceIndex,
+            nastranOrdering,
+            tractionDVNums,
         )
 
     def addPressureToComponents(self, compIDs, pressures, faceIndex=0):
```
eirikurj: I dont see `pressureDVNums` applied here (and elsewhere in the buckling class) like in the other problems. Any reason to skip? We should probably include it here as well.

tacs/pymeshloader.py
```diff
@@ -851,7 +851,7 @@ def getElementObjectForElemID(self, elemID, nastranOrdering=False):
         elemObj = self.elemObjects[elemObjNum]
         return elemObj
 
-    def createTACSAssembler(self, varsPerNode, massDVs):
+    def createTACSAssembler(self, varsPerNode, massDVs, numDVs, globalDVNums=None):
```
eirikurj: The `numDVs` input variable is documented, but does not seem to be used anywhere. Please remove.

tacs/problems/transient.py
```diff
@@ -842,6 +923,15 @@ def _updateAssemblerVars(self):
         c2 = self.getOption("RBEArtificialStiffness")
         tacs.elements.RBE2.setScalingParameters(c1, c2)
         tacs.elements.RBE3.setScalingParameters(c1, c2)
+        # Update design variables on all auxiliary elements
+        for auxElemObj in self.auxElems:
```
eirikurj: For time-dependent problems, do you think this process might be somewhat costly? If we consider inertial loads, then there is one aux element per local element, and looping in python over all elements at every timestep might be costly. Maybe I am overly cautious, but for performance it might be better to do the loop in the c++ layer instead. The structure here looks similar to the one in `TACSAssembler::setDesignVars`.

tacs/pytacs.py
```diff
@@ -821,17 +918,52 @@ def elemCallBack(dvNum, compID, compDescript, elemDescripts,
...
         self.tempVec = self.assembler.createVec()
 
+    def _setGlobalDVValues(self, vec, dv_nums, dv_vals):
+        """Insert global DV values into a distributed design vec."""
+        if not dv_nums:
+            return
+        vec.setValues(
+            np.array(dv_nums, dtype=np.intc),
+            np.array(dv_vals, dtype=self.dtype),
+            op=tacs.TACS.INSERT_NONZERO_VALUES,
```
eirikurj: Does this not skip all then all zero values? For example, setting `addGlobalDV("m", 500.0, lower=0.0, upper=2000.0)` the lower bound is not updated and stays as the default -1e20? 
We should probably use the `INSERT_VALUES`  and only have the root proc set the values (and drop the `begin/end`). Please check.

src/elements/TACSTraction3D.cpp
```diff
@@ -31,6 +32,13 @@ TACSTraction3D::TACSTraction3D(int _varsPerNode, int _faceIndex,
...
+  if (_tracDVNums) {
+    memcpy(tracDVNums, _tracDVNums, varsPerNode * sizeof(int));
+  } else {
+    for (int i = 0; i < varsPerNode; i++) {
+      tracDVNums[i] = -1;
+    }
+  }
 }
 
 TACSTraction3D::TACSTraction3D(
```
eirikurj, line 40: In the second constructor I dont see the `tracDVNums` being initialized. Maybe add a small helper to initialize in both?

src/elements/TACSTraction2D.cpp
```diff
@@ -31,6 +32,13 @@ TACSTraction2D::TACSTraction2D(int _varsPerNode, int _faceIndex,
...
+  if (_tracDVNums) {
+    memcpy(tracDVNums, _tracDVNums, varsPerNode * sizeof(int));
+  } else {
+    for (int i = 0; i < varsPerNode; i++) {
+      tracDVNums[i] = -1;
+    }
+  }
 }
 
 TACSTraction2D::TACSTraction2D(
```
eirikurj, line 40: In the second constructor I dont see the `tracDVNums` being initialized. Maybe add a small helper to initialize in both?

src/elements/shell/TACSBeamInertialForce.h
```diff
@@ -65,20 +71,57 @@ class TACSBeamInertialForce : public TACSElement {
   }
 
   int getDesignVarNums(int elemIndex, int dvLen, int dvNums[]) {
-    return con->getDesignVarNums(elemIndex, dvLen, dvNums);
+    int num = con->getDesignVarNums(elemIndex, dvLen, dvNums);
```
eirikurj, line 76: For these methods, `getDesignVarNums, getDesignVar,s setDesignVars, getDesignVarRange`, the loops are essentially identical across many if not all the element implementations, with the DV being the only variation. I think we should try to refactor these into something more general that can be reused when possible instead of adding these loops all over. This way these functions could be much cleaner and consistent. I have not fully thought this through but what I had in mind is something like 
```
int getDesignVarNums(int elemIndex, int dvLen, int dvNums[]) {
	int num = con->getDesignVarNums(elemIndex, dvLen, dvNums);
	TacsAppendLoadDVNums(3, omegaDVNums, dvLen, dvNums, &num);
	TacsAppendLoadDVNums(3, rotCenterDVNums, dvLen, dvNums, &num);
	return num;
}
```
where the implementation is basically the loop etc. Same pattern (or similar) for the other functions

src/elements/shell/TACSBeamInertialForce.h
```diff
@@ -65,20 +71,57 @@ class TACSBeamInertialForce : public TACSElement {
...
   }
 
   int getDesignVarRange(int elemIndex, int dvLen, TacsScalar lb[],
                         TacsScalar ub[]) {
-    return con->getDesignVarRange(elemIndex, dvLen, lb, ub);
+    int num = con->getDesignVarRange(elemIndex, dvLen, lb, ub);
+    for (int i = 0; i < 3; i++) {
+      if (inertiaVecDVNums[i] >= 0) {
+        if (num < dvLen) {
+          lb[num] = -1e20;
```
eirikurj: This `1e20` value is all over the place. We should probably define this a constant. See my refactoring comment above, but that would mitigate this by having only one location where these appear. Would stil be nice to define them somewhere.

tacs/problems/base.py
```diff
@@ -359,7 +363,36 @@ def _addLoadToRHS(self, Frhs, Fapplied):
...
+                raise self._TACSError(
+                    "Design variable numbers {} were not added as global DVs. "
+                    "Global DV nums registered within the assembler: {}.".format(
+                        bad.tolist(), sorted(self.globalDVNums)
+                    )
+                )
+        return dvNums
+
+    def _addTractionToComponents(
+        self, auxElems, compIDs, tractions, faceIndex=0, tractionDVNums=None
```
eirikurj, line 428: Maybe I missed this in the tests but I dont see any integration test using `tractionDVNums`

tacs/problems/buckling.py
```diff
@@ -610,7 +634,7 @@ def addPressureToElements(
             self.auxElems, elemIDs, pressures, faceIndex, nastranOrdering
         )
 
-    def addInertialLoad(self, inertiaVector):
+    def addInertialLoad(self, inertiaVector, inertiaVecDVNums=None):
```
eirikurj, line 661: Maybe I missed this, but I could not find any buckling tests that added any of the new load DVs.

eirikurj (commented):
Overall looks good, but please see my comments.

src/elements/shell/TACSBeamInertialForce.h
timryanb: Let's handle this in a seprate PR @eirikurj

tacs/problems/static.py
timryanb: Moved to a sub-routine in BaseProblem that is no called in each problems pre-solve  method

tacs/problems/transient.py
timryanb: Element loop moved into c++

src/elements/shell/TACSBeamInertialForce.h
timryanb: Added in a TACS_LARGE_DV_BOUND constant and replaced explicit 1e20 entries in auxelem dvs with new constants. There's probably other locations where this change can be made in the constitutive classes, but that can be done in a separate PR

tacs/problems/base.py
timryanb: Added two new subtests in test_shell_plate_buckling_axial.py and test_shell_plate_partitioned.py

tacs/problems/buckling.py
timryanb: Added a new subtests in test_shell_plate_buckling_axial.py

src/TACSAuxElements.cpp
```diff
@@ -141,29 +146,49 @@ int TACSAuxElements::getAuxElements(TACSAuxElem **_elems) {
...
+  TACSAuxElements instance currently attached to the assembler.
+
+  input:
+  dvs:  the design variable vector
 */
-void TACSAuxElements::getDesignVars(int numDVs, TacsScalar dvs[]) {
-  for (int i = 0; i < num_elements; i++) {
-    aux[i].elem->getDesignVars(i, numDVs, dvs);
+void TACSAuxElements::setDesignVars(TACSBVec *dvs) {
+  if (num_elements == 0) {
```
eirikurj: I think this could be a minor bug here if we return before `beginDistributeValues()` which is a collective call.  If rank 0 has no external DVs it returns and its distribute posts nothing, but every other rank must receive from it. So if rank 0 has no aux elements but other ranks do, they will be stuck there. If we want the early `return`, this should probably appear after the collective calls.

tacs/pytacs.py
```diff
@@ -821,17 +918,84 @@ def elemCallBack(dvNum, compID, compDescript, elemDescripts,
...
         self.xub = self.assembler.createDesignVec()
         self.xlb = self.assembler.createDesignVec()
         self.assembler.getDesignVarRange(self.xlb, self.xub)
+        # Overwrite with any global DV bounds specified by the user
+        global_dv_nums_ub = []
+        global_dv_vals_ub = []
+        global_dv_nums_lb = []
+        global_dv_vals_lb = []
+        for dv_info in self.globalDVs.values():
+            if dv_info["upperBound"] is not None:
```
eirikurj: The `is not None` filters mean a global DV with default bounds is never added to either list, so nothing writes its entry. `getDesignVarRange` on line 933 only covers DVs held by regular elements, and the aux elements that consume load DVs don't exist yet, so a global DV consumed only by aux loads keeps the zero from `createDesignVec()` in both bounds:

```
unbounded  dv=0  x0=1e+07  lb=0  ub=0        in-bounds=False
bounded    dv=1  x0=1e+07  lb=1  ub=2e+07    in-bounds=True
element    dv=3  x0=0.01   lb=0  ub=10       in-bounds=True
```

The initial value ends up outside its own bounds, and now propagates to every problem and constraint via the new `setDesignVarRange`. Both new tests demonstrate this `addGlobalDV("pressure", P), addGlobalDV("traction", trac)`, no bounds, but assert nothing about bounds. For example, 

```
dv = fea_assembler..addGlobalDV("pressure", 100e5)   # no bounds
fea_assembler.initialize()
lb, ub = fea.getDesignVarRange()
assert ub[dv] >= 100e5                     # fails as its currently 0.0
```

Not sure what the best approach is here, but maybe it is to apply the unbounded default only to global DVs that no regular element holds, determinable at `initialize()` by unioning `elem.getDesignVarNums(i)` over local elements and allgathering. We should also update the tests to test that bounds are properly added.

tests/integration_tests/test_multi_rectangular_beam.py
```diff
@@ -96,11 +97,18 @@ def elem_call_back(
...
+        assert fea_assembler.getGlobalDVNums() == [0, 1, 2]
+        assert fea_assembler.getTotalNumGlobalDVs() == 3
+
         # Set up constitutive objects and elements
         fea_assembler.initialize(elem_call_back)
 
         grav_prob = fea_assembler.createStaticProblem("gravity")
-        grav_prob.addInertialLoad([-10.0, 3.0, 5.0])
+        grav_prob.addInertialLoad(g_vec, inertiaVecDVNums=gDV)
+        grav_prob.getJacobian()
```
eirikurj: Is this line needed? Does not seem to do anything here. I suggest removing or adding some asserts.

eirikurj (commented):
Looks good, just a couple of other followup comments.

src/TACSAuxElements.cpp
timryanb: Good catch, removed the early return entirely so the collective distribute calls run on every rank (the loops below are natural no-ops with zero elements). Verified with a 2-rank repro that deadlocked before the fix.

tacs/pytacs.py
timryanb: Implemented as you suggested: at initialize, DV nums held by regular elements are unioned across ranks, and global DVs outside that set with unspecified bounds get ±TACS_LARGE_DV_BOUND defaults (element-held DVs keep the element-reported range). Added unit tests for the unheld, one-sided, and partially-held-array cases, and bounds assertions to both plate integration tests.

tests/integration_tests/test_multi_rectangular_beam.py
timryanb: removed

eirikurj (approved):
Looks good! Thanks for addressing my comments. Approving, but not merging in case @A-CGray intends to review this as PR as well given the size of it. If not you can go ahead and merge.

PR #468, opened 2026-07-24
Alternative implementation for constitutive model design variable groups

A-CGray:
#462 adds new methods to pyTACS that produce user-friendly dictionaries containing the values and indices for all design variables associated with each component in a model. Behind the scenes that was enabled by implementing 6 new methods in every constitutive class:

- `getNumDesignVarGroups`
- `getDesignVarGroupName`
- `getDesignVarGroupSize`
- `isDesignVarGroupScalar`
- `getDesignVarGroupValues`
- `getDesignVarGroupNums`

This PR provides the same public API with a new implementation that reduces the amount of code that needs to be implemented in each constitutive model.

Information about a constitutive model's design variable groups are now stored in a "table" (really a `std::vector<DesignVarGroup>`), each constitutive model populates this table using these two methods:

```cpp
void registerScalarDesignVarGroup(const char *name, TacsScalar *value, int *dvNum);
void registerArrayDesignVarGroup(const char *name, int size, TacsScalar *values, int *dvNums);
```

The six methods above, that used to have to be implemented by every constitutive model, are now non-virtual methods of the base `TACSConstitutive` class.

### DIFF
--- a/src/constitutive/TACSConstitutive.cpp
+++ b/src/constitutive/TACSConstitutive.cpp
@@ -135,3 +135,76 @@ void TACSConstitutive::getFailureEnvelope(
   delete[] y_strain;
   delete[] C;
 }
+
+void TACSConstitutive::registerScalarDesignVarGroup(const char *name,
+                                                    TacsScalar *value,
+                                                    int *dvNum) {
+  DesignVarGroup group;
+  group.name = name;
+  group.size = 1;
+  group.isScalar = true;
+  group.values = value;
+  group.dvNums = dvNum;
+  designVarGroups.push_back(group);
+}
+
+void TACSConstitutive::registerArrayDesignVarGroup(const char *name, int size,
+                                                   TacsScalar *values,
+                                                   int *dvNums) {
+  DesignVarGroup group;
+  group.name = name;
+  group.size = size;
+  group.isScalar = false;
+  group.values = values;
+  group.dvNums = dvNums;
+  designVarGroups.push_back(group);
+}
+
+int TACSConstitutive::getNumDesignVarGroups() {
+  return static_cast<int>(designVarGroups.size());
+}
+
+const char *TACSConstitutive::getDesignVarGroupName(int groupIndex) {
+  if (groupIndex >= 0 &&
+      groupIndex < static_cast<int>(designVarGroups.size())) {
+    return designVarGroups[groupIndex].name;
+  }
+  return NULL;
+}
+
+int TACSConstitutive::getDesignVarGroupSize(int groupIndex) {
+  if (groupIndex >= 0 &&
+      groupIndex < static_cast<int>(designVarGroups.size())) {
+    return designVarGroups[groupIndex].size;
+  }
+  return 0;
+}
+
+bool TACSConstitutive::isDesignVarGroupScalar(int groupIndex) {
+  if (groupIndex >= 0 &&
+      groupIndex < static_cast<int>(designVarGroups.size())) {
+    return designVarGroups[groupIndex].isScalar;
+  }
+  return true;
+}
+
+void TACSConstitutive::getDesignVarGroupValues(int groupIndex,
+                                               TacsScalar values[]) {
+  if (groupIndex >= 0 &&
+      groupIndex < static_cast<int>(designVarGroups.size())) {
+    const DesignVarGroup &group = designVarGroups[groupIndex];
+    for (int ii = 0; ii < group.size; ii++) {
+      values[ii] = group.values[ii];
+    }
+  }
+}
+
+void TACSConstitutive::getDesignVarGroupNums(int groupIndex, int dvNums[]) {
+  if (groupIndex >= 0 &&
+      groupIndex < static_cast<int>(designVarGroups.size())) {
+    const DesignVarGroup &group = designVarGroups[groupIndex];
+    for (int ii = 0; ii < group.size; ii++) {
+      dvNums[ii] = group.dvNums[ii];
+    }
+  }
+}
--- a/src/constitutive/TACSConstitutive.h
+++ b/src/constitutive/TACSConstitutive.h
@@ -131,7 +131,7 @@ class TACSConstitutive : public TACSObject {
 
     @return The number of design variable groups
   */
-  virtual int getNumDesignVarGroups();
+  int getNumDesignVarGroups();
 
   /**
     Get the name of a design variable group
@@ -139,7 +139,7 @@ class TACSConstitutive : public TACSObject {
     @param groupIndex The index of the design variable group
     @return The group name, or NULL if groupIndex is out of range
   */
-  virtual const char *getDesignVarGroupName(int groupIndex);
+  const char *getDesignVarGroupName(int groupIndex);
 
   /**
     Get the number of entries in a design variable group
@@ -147,7 +147,7 @@ class TACSConstitutive : public TACSObject {
     @param groupIndex The index of the design variable group
     @return The number of entries in the group, or 0 if out of range
   */
-  virtual int getDesignVarGroupSize(int groupIndex);
+  int getDesignVarGroupSize(int groupIndex);
 
   /**
     Is this design variable group a scalar quantity?
@@ -159,7 +159,7 @@ class TACSConstitutive : public TACSObject {
     @param groupIndex The index of the design variable group
     @return True if the group is a scalar quantity
   */
-  virtual bool isDesignVarGroupScalar(int groupIndex);
+  bool isDesignVarGroupScalar(int groupIndex);
 
   /**
     Get the values of all entries in a design variable group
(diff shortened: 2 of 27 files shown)


A-CGray:
@timryanb following up on your question about what happens if a constitutive model doesn't register one of its design variable groups. Currently the way this is caught is through tests, specifically `assertGroupsConsistent` which checks that the design variable values and numbers returned by the new API match those returned by the old API (`getDesignVarNums/getDesignVars`). At runtime, if a design variable group is missing, it's data just won't be contained in the dictionaries returned by `getComponentDesignVars/getComponentDesignVarNums`.

Obviously the weakness here is that this only catches an unregistered DV group in a constitutive model if we have a `DVGroupTestCase` for that constitutive model in `test_dv_groups.py`.

I think the only way to 100% guarantee that we avoid missing design variable groups in future would be to make this design variable group API a more core part of the constitutive model. At a minimum that'd mean re-implementing `getDesignVarNums/getDesignVars` to use the design variable group data. At the other end of the spectrum, we could change the constitutive models so that design variable information is only stored in the `designVarGroups` table, i.e we'd get rid of all the class attributes that currently store the design variable values etc in each class. There would be some big benefits to that, but it's a way bigger change.

FYI, I think this issue is present in both this implementation and the one in #462

src/constitutive/TACSConstitutive.cpp
```diff
@@ -135,3 +135,76 @@ void TACSConstitutive::getFailureEnvelope(
   delete[] y_strain;
   delete[] C;
 }
+
+void TACSConstitutive::registerScalarDesignVarGroup(const char *name,
```
eirikurj, line 139: The `registerScalarDesignVarGroup` and `registerArrayDesignVarGroup` do most of the same things. The only thing that differentiates these is `size` and `isScalar`. We should have a third routine that does all the common tasks once, and these just call that routine.

eirikurj, line 139: I think we should add some minimal checks here. For example, we can check 
- if the same DV name has already been added or not
- that the size > 0 
- that passed pointers are not NULL
- (check the values of `dvNums >= -1`?)

Not entirely sure what the action should be though if any of these happen. That is, should we allow failed registration and print a warning or just full stop break. See my other comment on checking consistency.

src/constitutive/TACSConstitutive.h
```diff
@@ -680,8 +683,43 @@ class TACSConstitutive : public TACSObject {
...
+  */
+  void registerArrayDesignVarGroup(const char *name, int size,
+                                   TacsScalar *values, int *dvNums);
+
  private:
   static const char *constName;
+
+  // A registered design variable group: a named quantity backed by live
+  // storage in the subclass. Scalar groups have size 1 and isScalar true.
+  struct DesignVarGroup {
```
eirikurj, line 713: Are we worried about objects being copied? Since these are pointers, the copy would refer to its original. Maybe we should disallow a copy and assignment constructor to prevent this happening?
Something like 
```
TACSConstitutive(const TACSConstitutive &) = delete;
TACSConstitutive &operator=(const TACSConstitutive &) = delete;
```

src/constitutive/TACSConstitutive.h
eirikurj, line 1: I thought a bit about the self-checking. This relates to the implementation in #462 where `_checkDVGroupsImplemented` only checks the number of active DVs, but not the order.

What about adding a member function here that checks against the DVs array? This would catch a group registered with e.g. the wrong `dvNums` pointer or the wrong registration order. This could be used directly in testing with `test_dv_groups.py` base assertion for every class and later from `_checkDVGroupsImplemented`

This could look something like the below. This might need some work, but hopefully the idea comes across.

```
/*
  Check that the design variable group table is consistent with the legacy
  DV API: the active entries (dvNum >= 0) of every group, concatenated in
  registration order, must exactly reproduce getDesignVarNums().
*/
bool TACSConstitutive::checkDesignVarGroupConsistency(int elemIndex) {
  // Get the active DV numbers from the legacy API
  int numDVs = getDesignVarNums(elemIndex, 0, NULL);
  std::vector<int> legacyNums(numDVs);
  getDesignVarNums(elemIndex, numDVs, legacyNums.data());

  int index = 0;  // Position in the legacy DV array
  for (size_t g = 0; g < designVarGroups.size(); g++) {
    const DesignVarGroup &group = designVarGroups[g];
    for (int ii = 0; ii < group.size; ii++) {
      if (group.dvNums[ii] >= 0) {
        if (index >= numDVs || group.dvNums[ii] != legacyNums[index]) {
          fprintf(stderr,
                  "TACSConstitutive: in %s, active entry %d of group \"%s\" "
                  "(DV num %d) does not match getDesignVarNums() at position "
                  "%d\n",
                  getObjectName(), ii, group.name, group.dvNums[ii], index);
          return false;
        }
        index++;
      }
    }
  }
  if (index != numDVs) {
    fprintf(stderr,
            "TACSConstitutive: in %s, design variable groups cover %d active "
            "DVs but getDesignVarNums() reports %d\n",
            getObjectName(), index, numDVs);
    return false;
  }
  return true;
}
```

src/constitutive/TACSConstitutive.h
```diff
@@ -680,8 +683,43 @@ class TACSConstitutive : public TACSObject {
...
+    const char *name;
+    int size;
+    bool isScalar;
+    TacsScalar *values;
+    int *dvNums;
+  };
+
+  // Design variable groups registered by the subclass, in registration order.
+  std::vector<DesignVarGroup> designVarGroups;
```
eirikurj, line 722: This would be somewhat of a code optimization, but it might make sense to have each child class preallocate the size of the vector with `designVarGroups.reserve(numGroups)` since we know the size upfront. Since this variable is private here, we might need to have a protected helper function e.g. `reserveDesignVarGroups(int n)` that does this. See what you think.

src/constitutive/TACSConstitutive.cpp
```diff
@@ -135,3 +135,76 @@ void TACSConstitutive::getFailureEnvelope(
...
+  group.dvNums = dvNums;
+  designVarGroups.push_back(group);
+}
+
+int TACSConstitutive::getNumDesignVarGroups() {
+  return static_cast<int>(designVarGroups.size());
+}
+
+const char *TACSConstitutive::getDesignVarGroupName(int groupIndex) {
+  if (groupIndex >= 0 &&
```
eirikurj, line 168: The same range check `if (groupIndex >= 0 && groupIndex < static_cast<int>(designVarGroups.size()))` appears in all functions. I think we could refactor and encapsulate this check into a helper function that just returns a pointer to the group. Would make all functions cleaner and more maintainable. Something like the following prototype (probably needs adjustment)
```
const DesignVarGroup *getGroup(int groupIndex) {
  if (groupIndex >= 0 && groupIndex < static_cast<int>(designVarGroups.size())) {
    return &designVarGroups[groupIndex];
  }
  return NULL;
}
```
Then we could do something like 
```
const char *TACSConstitutive::getDesignVarGroupName(int groupIndex) {
  const DesignVarGroup *group = getGroup(groupIndex);
  if (group) {
    return group->name;
  }
  return NULL;
}
```

src/constitutive/TACSConstitutive.cpp
eirikurj, line 139: The new functions added are not documented. We should probably add minimum docs.

src/constitutive/TACSConstitutive.cpp
eirikurj, line 168: I think it might make sense to have one function that gets all properties at once, but this would minimize the python-c++ crossing, getting all members at once.

eirikurj (commented):
I do like this design much better than what is in #462 and think it is cleaner and a lot less code. Tried to be mindful when scaling this up, as mentioned in the last meeting. The one thing here is the addition of the table, which might increase memory usage for lots of constitutive objects, but does not seem to be the case for the #462 approach. I think regardless of the memory hit here, this PR is the better approach. Overall this looks good, just a few comments.

src/constitutive/TACSConstitutive.cpp
timryanb: Is there a reason we can't combine these into one function and require the user to specify size=1 for scalar? Having two separate methods feels awkward

timryanb: We do't need to restrict users to `dvNum >= -1`. I don't think TACS cares what negative value a dv takes, it just checks if `dvNum >=0` to determine if its active. -1 is just a convention, but any negative value should be allowed to work.

A-CGray: The reason for separate methods is that there is a difference between a scalar DV and an array DV with a size of 1. For example, you might have a constitutive class that has ply fraction DVs, but you only use a single ply.

A-CGray: The docstrings are in the header file

timryanb: @A-CGray I think I'm missing something, what does having a specific class for scalar dvs buy us that having one general class that supports size one arrays doesn't? Could a thickness DV not just be registered as a length 1 dv array?

PR #469, opened 2026-07-28
Feature beam element methods

timryanb:
## Purpose

Brings `TACSBeamElement` up to the same analytic-derivative parity as `TACSShellElement`: an analytic `addJacobian`, analytic `TACS_GEOMETRIC_STIFFNESS_MATRIX` support, and analytic matrix-sensitivity (`addMatDVSensInnerProduct`/`addMatXptSensInnerProduct`/`getMatSVSensInnerProduct`) overrides.

This is implemented by extending the `src/elements/a2d/` library with second-order (Hessian-carrying) `hforwardr than porting shell'shand-coded congruence-transform kernels.

**Scope landed:**                                                                   
 - `TACSBeamElement::addJacob built as a second-order(`hforward`/`hreverse`) extension of the existing `addResidual` A2D graph, one sweep per state DOF.
- `TACSBeamNonlinearModel::evalStrain/Sens/Deriv/Hessian` completed (was previouscommented out/incomplete), uFFNESS_MATRIX`.
- `TACSBeamElement::getMatType(TACS_GEOMETRIC_STIFFNESS_MATRIX)` — analytic port (previously returned all zer
- `addMatDVSensInnerProduct`/`addMatXptSensInnerProduct`/`getMatSVSensInnerProdanalytic for `TACS_STIFFNESSall three director classes);`getMatSVSensInnerProduct`'s `TACS_GEOMETRIC_STIFFNESS_MATRIX` branch is analyt`TACSLinearizedRotation`.
- `TACSBeamElement::getAverageStresses` — new override (was a silent no-op, a r`Assembler.getAverageStresseomponents).
- `addPointQuantityXptSens`'s `TACS_ELEMENT_MOMENT_OF_INERTIA` branch — closed analytically (was the one retype).
- `TACSDirector.h` gains `addDirectorHessianProduct`/`addDirectorHessianRefNormalSens`hooks, plus a new `(d01,d02)ure(`TacsBeamAddCrossDirectorJacobian`) and a new                                        `TACSBeamElementModel::addCoure — both genuinely newderivations with no shell precedent, needed to assemble `addJacobian`'s tying-strain  and dual-director contributi
- `src/elements/a2d/` (`a2dobjs.h`, `a2dvecops.h`, `a2dmatops.h`) extended in place   with `xp`/`xh`, `valuep`/`va `hforward`/`hreverse` methods on the 6 state-dependent op-structs the beam's residual graph uses, each with a       standalone op-level complex-/a2d/tests/`.
- Three previously `@unittest.SkipTest`'d matrix-sens tests in `test_beam_element.py` are un-skipped; new `test_ben Euler-column`TACSLinearBuckling` integration test (`test_beam_column_buckling.py`) added.
- Dead, never-finished shellACSBeamUtilities.h` deleted.

**Explicitly deferred, not a(documented at the code sites, still forwarded to the base FD/complex-step fallback):**
- `addMatDVSensInnerProduct`s`TACS_GEOMETRIC_STIFFNESS_MATRIX` branches (attempted once, found insufficient — see
the commit documenting the f from-scratch pass, not aguess).
- Analytic `TACS_GEOMETRIC_Sfor the nonlinear directorclasses (`TACSQuadraticRotation`/`TACSQuaternionRotation`) — one attempt (G1) found a
genuine second obstruction (the recipe didn't account for) and was reverted in favor of documenting it; the follow-on branches (G2–G4) were
consequently not attempted rfied foundation.

Both gaps leave existing fallback, as before this PR) — no regression, just unfinished analytic coverage, flagged for a follow-up.


A-CGray:
@timryanb I see that Claude made some changes to the A2D source code in this PR. My understanding is that the A2D code in this repo is quite out of date with the actual a2d repo (https://github.com/smdogroup/a2d) and I don't really like the idea of maintaining a separate version within the TACS source. I think we should port the A2D changes made here to https://github.com/smdogroup/a2d. Then in terms of including A2D in TACS we could either:

- Keep the current approach, copy the latest A2D into the TACS source
- Include A2D in the tacs repo as a git submodule
- Setup TACS to rely on A2D as an external dependency, like we do with METIS, AMD, and tecio

timryanb:
> @timryanb I see that Claude made some changes to the A2D source code in this PR. My understanding is that the A2D code in this repo is quite out of date with the actual a2d repo (https://github.com/smdogroup/a2d) and I don't really like the idea of maintaining a separate version within the TACS source. I think we should port the A2D changes made here to https://github.com/smdogroup/a2d. Then in terms of including A2D in TACS we could either:
> 
> * Keep the current approach, copy the latest A2D into the TACS source
> * Include A2D in the tacs repo as a git submodule
> * Setup TACS to rely on A2D as an external dependency, like we do with METIS, AMD, and tecio

Fair point. I'd rather avoid adding new non-python external dependencies that users have to install. So my preference would be for option 1 or 2. I'm not familiar with submoduling in git, but that might be a good idea for including the library w/ the tacs source

PR #471, opened 2026-07-31
[Bugfix] Make Tsai-Wu failure criterion equivalent to von Mises for isotropic materials

A-CGray:
TACS constitutive models that use the `TACSOrthotropicPly` class (e.g `TACSBladeStiffenedShellConstitutive`) are in theory compatible with isotropic materials because the modified Tsai-Wu criterion we use can be made mathematically equivalent to the von Mises stress. However, this relies on two things that are not true in the current source code:

- `F12*=-1/2`, [currently it's hard-coded to zero](https://github.com/smdogroup/tacs/blob/8daad07bc6bb0ba794a4561685fb7621a5f6e854/src/constitutive/TACSMaterialProperties.cpp#L552-L575), which is valid for orthotropic materials
- The shear strength should by $\sigma_y/\sqrt{3}$, currently [it's set to](https://github.com/smdogroup/tacs/blob/8daad07bc6bb0ba794a4561685fb7621a5f6e854/src/constitutive/TACSMaterialProperties.cpp#L52) $\sigma_y \times \sqrt{3}$ for isotropic materials.

This PR does this following:
- Ensures Tsai-Wu matches von Mises by default when using an isotropic material by:
  - Fixing the shear strength bug in `TACSMaterialProperties`
  - Setting `F12*=-1/2` in `TACSOrthotropicPly` when the associated material is isotropic, the existing default of zero is maintained for orthotropic materials
- Updates integration test reference KSFailure values
- Adds tests verifying the equivalence of the two criteria for isotropic materials
- Adds a section to the docs explaining all of the failure criteria available through `TACSOrthotropicPly`


timryanb (approved):
Looks good to me. Should we make a bugfixx release @A-CGray?

A-CGray:
> Looks good to me. Should we make a bugfixx release @A-CGray?

Sure, I bumped the patch version here now.

PR #474, opened 2026-08-11
Restrict UCX transports in the conda test environment

timryanb:
The mpich variants abort in MPI_Init when UCX finds an RDMA device it cannot open, which has failed the linux-64 complex deploy since v3.12.2.


A-CGray (approved):
I have no idea what this does but I'm assuming it fixes some error that you ran into?

timryanb:
> I have no idea what this does but I'm assuming it fixes some error that you ran into?

Sorry, yes this fixes an issue that we were seeing with the post build tests for the complex MPICH linux build [here](https://github.com/smdogroup/tacs/actions/runs/31410998606) @A-CGray

PR #476, opened 2026-08-29
[CI debug] Surface the silent tecio solve failure on macOS

eirikurj:
Temporary debugging of the failing macOS CI, not for merge.

### DIFF
--- a/.github/workflows/unit_tests.yml
+++ b/.github/workflows/unit_tests.yml
@@ -129,6 +129,20 @@ jobs:
           echo "::group::solve probe: tecio + numpy>=2 (what CI actually asks for)"
           mamba install -c smdogroup -c conda-forge tecio "numpy>=2" --dry-run -y || true
           echo "::endgroup::"
+          # Force the channel so the solver must report WHY it cannot use smdogroup's
+          # tecio, instead of silently substituting the same-named conda-forge package.
+          echo "::group::solve probe: smdogroup::tecio + numpy>=2 (expect a conflict report)"
+          mamba install -c smdogroup -c conda-forge smdogroup::tecio "numpy>=2" --dry-run -y || true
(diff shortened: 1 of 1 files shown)


eirikurj:
Superseded for the actual fix by #478, which installs tecio into its own env. Keeping this open for the solver output until #478 is green.

eirikurj:
Closing. Fix proposed in #478.

**Root cause:** two different packages are named `tecio`. `smdogroup::tecio 2021.2` is the Tecplot C library f5totec needs; `conda-forge::tecio 2.0.6` is an unrelated pure-python reader with no headers. Because the workflow installed the bare name into the test env, the solver could choose between them on cost:

- `conda-forge::tecio`: 5 installs, 6 MB, no downgrades
- `smdogroup::tecio`: 14 installs, 2 removals, 11 build changes, 5 downgrades, 113 MB — because its `boost` run-dep pulls `libboost` -> `icu <76`, forcing removal of the LLVM 23 toolchain and downgrades of icu, libxml2 and libpsl

Neither is unsatisfiable, so nothing errored. The expensive branch got more expensive when conda-forge moved to libcxx/LLVM 23.1.0, the solver flipped to the cheap wrong package, and the failure surfaced only as `fatal error: 'TECIO.h' file not found` at compile time. This is why it broke with no change to this repo and no change to either tecio package.

**Fix (#478):** install tecio into its own env with the channel named explicitly, and verify the header and static lib exist before compiling. Confirmed green on macOS and Linux.

**Still worth doing:** rebuild `smdogroup::tecio` without the `boost` run dependency. It is the sole source of the expensive branch and is a build-time leak — #478's CI links `libtecio.a` from an env with no boost at all, so the static library plainly does not need it. See the maintainer note in #478.

